PR #11, opened 2017-05-31
Move backends into submodules, include initial equations and value assignments in flattening, and various other fixes

jbaayen:


pymola/ast.py
```diff
@@ -211,7 +211,7 @@ def __init__(self, **kwargs):
         self.max = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
         self.nominal = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
         self.value = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
-        self.fixed = False  # type: bool
+        self.fixed = Primary(value=False)  # type: Primary
```
jgoppert: what is the thinking here? I  would think that making it a primary is overkill if it is always going to be type bool?

jbaayen: The way the grammar is set up, any constant value, whether float, integer, or a boolean, gets handled as a Primary. Since we handle attributes in a generic way (lines 551-552 of parser.py), we would have to make a special exception for the attribute 'fixed' if it is not of the type Primary like its cousins.

jgoppert:
Please rebase on master now that I fixed the issue with testing and make sure the CI passes.

jgoppert:
Mind adding a unit test for simplify etc. Would be good to work towards 100% coverage.

jbaayen:
I've added unit tests now for simplify() and finished the refactor of the backend. It should be good to pull now.

jbaayen:
Incidentally, what is the casadi version installed on the CI node?
I'm trying to figure out why some of the unit tests succeed on my machine, but fail on travis.

jgoppert:
Nice, here is the casadi version, anaconda.org/conda-forge/casadi 3.1.1

jbaayen:
Ah! That may be the problem. The code requires casadi >= 3.2 now.

PR #12, opened 2017-06-29
Move backends into submodules, include initial equations and value assignments in flattening, and various other fixes

jbaayen:
This is the same merge request as before, just moved to a different branch.


jgoppert:
@jbaayen To get the unit test working you can modify this file: https://github.com/jgoppert/pymola/blob/master/create_conda_env.sh

Just remove the conda install casadi and switch to pip install casadi.

jbaayen:
All tests pass now. The problem was some non-determinism in the code due the use of regular dictionaries, as opposed to OrderedDict. The CasADi unit tests, on the other hand, were set up to expect a fixed certain ordering. This is fixed on both ends now: OrderedDict is used when appropriate, and the unit test checker is now independent of the order of the variable definitions.

jgoppert:
Great, looks good.

PR #14, opened 2017-06-30
Use distutils to compile CasADi models reliably on any platform

jbaayen:


jbaayen:
It turns out this needs further work. Closing for the moment.

I wonder if travis can be set up to run CI on our fork as well?

jgoppert:
@jbaayen @jackvreeken I gave you push access so you don't have to worry about setting up CI for your fork, just make sure  someone reviews major PR's, and we get all the unit tests passing before we merge it to master.

PR #23, opened 2017-10-03
WIP: Add support for for/else/when blocks.

jgoppert:
This is a work in progress to add support for various control structures we hadn't needed yet. My goal is to get when statements working for simulation of a model with discrete sampling.

# TODO
* I need to get the backend for sympy updated to support the new control structures.
* I broke casadi for loop generation, so I might need your help with that @jbaayen .
* I disabled cython as I was having trouble getting it running on my current environment, so I need to figure that out.


setup.py
```diff
@@ -21,7 +21,7 @@
 from setuptools.command.build_ext import build_ext
 from setuptools import Command
 
-from Cython.Build import cythonize
+# from Cython.Build import cythonize
```
jgoppert: Need to re-enable this once I get it working on my side.

pymola/parser.py
```diff
@@ -258,19 +278,33 @@ def exitStatement_component_function(self, ctx):
...
-            statements=[self.ast[s] for s in ctx.if_statement().statement()])
+    def exitStatement_if(self, ctx: ModelicaParser.Statement_ifContext):
+        """fall through to if_statement"""
+        self.ast[ctx] = self.ast[ctx.if_statement()]
 
-    def exitStatement_for(self, ctx):
+    def exitStatement_for(self, ctx: ModelicaParser.Statement_forContext):
+        """fall through to for_statement"""
+        self.ast[ctx] = self.ast[ctx.for_statement()]
+
```
jgoppert: Changes to if statement below.

pymola/Modelica.g4
```diff
@@ -355,27 +365,27 @@ for_index :
...
     'end' 'when'
     ;
 
 // B.2.6.15 ------------------------------------------------
 when_statement:
-    'when' expression 'then'
-        (statement ';')*
-    ('elsewhen' expression 'then'
-        (statement ';')*
+    'when' when_condition=expression 'then'
```
jgoppert: Added labels here for each section, when, elsewhen

pymola/parser.py
```diff
@@ -44,13 +44,13 @@ def class_node(self):
 
     # FILE ===========================================================
 
-    def enterStored_definition(self, ctx):
+    def enterStored_definition(self, ctx: ModelicaParser.Stored_definitionContext):
```
jgoppert: Adding the type hints here.

pymola/Modelica.g4
```diff
@@ -535,8 +545,8 @@ WS  :   [ \r\n\t]+ -> skip ; // toss out whitespace
 //=========================================================
 fragment Q_IDENT : '\'' ( Q_CHAR | S_ESCAPE)+;
 fragment NONDIGIT : [_a-zA-Z];
-fragment S_CHAR : [A-Za-z\u0000-\u00FF];
-fragment Q_CHAR : NONDIGIT | DIGIT | [!#$%&()*+,-./:;<>=?@[\]^{}|! ];
+fragment S_CHAR : [\u0000-\u00FF];
+fragment Q_CHAR : NONDIGIT | DIGIT | [!#$%&()*+,-./:;<>=?@[\]^{}| ];
```
jgoppert: This is a fix for a warning I was getting from antlr about redundant characters.

jackvreeken:
I'm a bit confused why the restructure into equation/statement blocks is needed. Wouldn't it also be possible to do it without adding this extra layer?

jgoppert:
@jackvreeken It allows you to sync up the conditions and statements a bit easier.

So block[0] executes if condition[0] holds.

Let me know if there is a better way to do it. It could probably be handled in the parser if you think the extra layer is too much work. It also cleans up the grammar a bit.

jgoppert:
The merge is a nightmare, I'll simplify this and rebase.

jackvreeken:
Ah, I think I get it. With the way things are, if there is more than one expression per branch, it is indeed impossible (as far as I can tell) to figure out what expression belongs in what branch. So your block approach makes a lot more sense.

I was wondering why we did not run into that before, but then I saw this in the CasADi back-end: 
`assert (len(tree.conditions) + 1 == len(tree.expressions))`
I guess we never needed more than one expression per branch yet...

jgoppert:
@jackvreeken This is rebased, so it would be good to start fixing the backends to support this.

jackvreeken:
Alright, I'll probably take a look sometime next week. Currently trying/fixing some other stuff in the CasADi backend, and I want to try and finish that first (but not necessarily push yet).

jackvreeken:
Will you add new small MRs on top of master, e.g. for adding the When statement? I can also try, but I'm not entirely sure about everything. There is something about function call arguments in the changes of this MR, of which I'm not sure what it is for. Furthermore, there are a whole lot of Sympy example (?) changes, which should probably be a separate commit/MR.

PR #26, opened 2017-11-15
Merge CasADi back-end changes

jackvreeken:


jackvreeken:
Coverage decreasing probably due to removing the Cython dependency.

jgoppert:
Let me know what I can do to help get this merged. Would like to get all of the recent PR's in so we are in sync.

jackvreeken:
I think it is actually OK. It involves some refactoring in the CasADi backend and the flattening process, but it has only made the code more adherent to the spec,  and comes with more test cases as well. The code coverage thing decreasing is because of Cython, but you knew that already. So I think we can go ahead an merge this, because all SymPy test cases still succeed as well.

PR #32, opened 2017-12-01
Remove unused files and cleanup AUTHOR list

### LINKED ISSUE #31: Clean up unneeded or obsolete files
I noticed some files/folders of which I'm not sure what they are for, or why they are there:

1. `.idea` - generally I avoid committing these to shared repositories, as people will have different preferences (and IDEs/editors). For example, it kept updating to Python 3.6 on mine (and a bunch of environment variables I needed in PyCharm), always cluttering my `git status`. I'd rather have it added to `.gitignore` to be honest.
2. `compiler.py` - I have to test if this still works, but if we keep it it should probably be moved to a folder `tools/` or something. Not something for the top level.
3. `create_conda_env.sh` - Is installation with conda that more involved that we need to include shell script? If a `pip install pymola` is not enough, maybe we should publish a conda package as well?
4. `wercker.yml` - Is wercker still used for something. It seems to build/publish docker images, although I am not sure what for.
5. `pymola/backends/casadi/AUTHORS` - Typically something placed at the top level as far as I know. But maybe other projects like Pymola (i.e. multiple back-ends with different people maintaining them) also do it this way... 
6. `doc/ModelicaSpec33.pdf` - Already out of date (3.4 is latest), so either we need to update it or remove it (if we don't want to keep updating it).

jgoppert:


jackvreeken:
"Closes #31" is not a very descriptive commit title, but otherwise looks good to me.

PR #37, opened 2018-01-08
fixes error when pymola attempts to sub an alg state for a parameter

vdwees:
eg. an equation where alg_state=param

### DIFF
--- a/src/pymola/backends/casadi/model.py
+++ b/src/pymola/backends/casadi/model.py
@@ -382,17 +382,19 @@ class Model:
             der_states = OrderedDict([(s.symbol.name(), s) for s in self.der_states])
             alg_states = OrderedDict([(s.symbol.name(), s) for s in self.alg_states])
             inputs = OrderedDict([(s.symbol.name(), s) for s in self.inputs])
+            parameters = OrderedDict([(s.symbol.name(), s) for s in self.parameters])
 
             all_states = OrderedDict()
             all_states.update(states)
             all_states.update(der_states)
             all_states.update(alg_states)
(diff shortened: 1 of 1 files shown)


jackvreeken:
I am a bit confused as to what `alg_states` are . Why can't they not be aliased with inputs and parameters?

```Modelica
model Test
  input Real x;
  parameter Real y = 3.0;
  Real z, w, v;
equation
  z = w * v;
  w = y;
  v = x
end Test;
```

I would expect that we end up with one state/alg_state `z`, with the `w` and `v` aliased to a parameter and input respectively.

vdwees:
Yes, alg_states can be substituted with parameters. But if we don't add the parameters to the allstates list, then when they become the canonical symbol there is a KeyError. In your example, if y does not have a value defined, e.g:
```modelica
model Test
  input Real x;
  parameter Real y;
  Real z, w, v;
equation
  z = w * v;
  w = y;
  v = x
end Test;
```
Then when w gets aliased to y, there will be an error because
1. y was not substituted with a number, so it is still symbolic
1. y's symbol is not in allstates, so we can't substitute w for y in the equations.

This change fixes this issue. For whatever reason, we forgot about them when this loop was written.

jackvreeken:
It also failed on my example by the way. I had to skim through the code again to remember what it was doing, but then it was immediately clear you are right. I was a bit confused by the `do_not_eliminate` variable in your changes earlier, thinking that meant we would not alias parameters anymore. But it only means that we would remove the alg_state if it found an alias, and never the parameter.

Issue #39, opened 2018-01-10
overly complicated equations limit equation optimization 


vdwees:
There are opportunities for equation elimination that are missed because of overly complicated equations. For example:
```
(-((Aarkanaal_1.Q[1]-Aarkanaal_1.Q[2])/3652.25))
```
If this equation were simplified to
```
Aarkanaal_1.Q[1]-Aarkanaal_1.Q[2]
```
the alias detection algorithm would detect and eliminate the equation. This would improve performance and reliability of the model. We should probably factor equations into an expression and a multiplier, and throw away multipliers.

PR #41, opened 2018-01-11
implement a simple factor and simplify algorithm for equations

vdwees:
@jackvreeken What do you think of this solution to #39 ?

### DIFF
--- a/src/pymola/backends/casadi/model.py
+++ b/src/pymola/backends/casadi/model.py
@@ -375,6 +375,41 @@ class Model:
                         [value] = ca.substitute([value], symbols, values)
                         setattr(variable, attribute, value)
 
+        if options.get('factor_and_simplify_equations', False):
+            # Operations that preserve the equivalence of an equation
+            # TODO: There may be more, but this is the most frequent set
+            unary_ops = ca.OP_NEG, ca.OP_FABS, ca.OP_SQRT
+            binary_ops = ca.OP_MUL, ca.OP_DIV
+
+            # Recursive factor and simplify function
+            def factor_and_simplify(eq):
+                # These are ops that can simply be dropped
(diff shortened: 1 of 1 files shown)


jackvreeken:
The first thing I thought of when you opened the issue was to use derivatives to check equality. Take the derivative of the equation to its first variable, and compare that to the derivative to the second variable. That would work for the example you gave in the issue, but probably won't when `sq()` / `exp()` etc. are part of the equation.

Old but related discussion over at CasADi: https://github.com/casadi/casadi/issues/1540

jackvreeken:
Another idea:

```Python
vars = ca.symvar(eq)
if len(vars) == 2:
    if ca.substitute(eq, vars[0], vars[1]).is_zero():
        # They are aliases!
        make_vars_aliases()
```

vdwees:
As you mentioned, it is slow and expensive. But I am also not sure it does what we want either- my objective for this algorithm is to simplify equations, not detect aliases. For example, `(0.48 *(A + B))/200` should be simplified to `A+B`. Whether we alias `A` to `B` or not, the simplified equation is a huge improvement from a numerical point of view.

jackvreeken:
Ah, I get what you mean. Your implementation looks like a good start. Does it also work with something like ```3600 * x + 3600 * y```? 

I would definitely like to check with @jgillis eventually, but if you already know now that it saves a lot of computations, I wouldn't be against merging what you made now.

vdwees:
No it does not. But it isn't common for that to happen unless the equation is written like that in the .mo
Still, it would be nice to catch that too.

vdwees:
Lets make the default behaviour to not do this simplification and then merge it.

vdwees:
@jackvreeken Can you squash commits when merging?

PR #46, opened 2018-01-25
Add if equation block support.

jgoppert:
@jackvreeken This is the first step towards supporting blocks. I just focused on the if equation to get the minimal patch so it would be easy to review.


test/gen_casadi_test.py
```diff
@@ -327,9 +329,10 @@ def test_if_else(self):
         ref_model.parameters[0].value = 10
         ref_model.equations = [
             y1 - ca.if_else(x > 0, 1, 0) * y_max,
-            ca.if_else(x > 1, ca.vertcat(y3 - 100, y2 - y_max),
-                       ca.if_else(x > 2, ca.vertcat(y3 - 1000, y2 - y_max - 1),
-                                  ca.vertcat(y3 - 10000, y2)))]
+            ca.if_else(x > 1, ca.vertcat(y2 - y_max, y3 - 100),
+                       ca.if_else(x > 2, ca.vertcat(y2 - y_max - 1, y3 - 1000),
+                                  ca.vertcat(y2, y3 - 10000)))
+        ]
```
jgoppert: This required a reordering the equations, the way I ordered it within the blocks is more logical, matches the source, so fixed it here to match.

jackvreeken:
Great to see you explore the CasADi backend! 

I took a stab at your branch sympy branch earlier today, rebasing it and splitting it up in smaller chunks: https://github.com/pymola/pymola/commits/condition-block-rebase .  I got more or less to the same point as you did here, and am now fixing the IfStatement (the last WIP commit, but without the `runme.py` file which is just for easy debugging). I'll finish that up tomorrow morning, and then send you a PR.

I agree about the ordering of the equations. I encountered and fixed that as well.

jgoppert:
@jackvreeken OK, awesome.

PR #48, opened 2018-02-07
get_derivative() now handles case where symbol is already indexed

vdwees:
with this pr, when a requesting the derivative of an indexed symbol, the slice is "passed through" the derivative operation and applied to the derivative symbol. e.g:
```
get_derivative(MX(a[2]))
```
now returns

```
MX(der(a)[2])
```
Previously, the behavior of get_derivative() was dramatically incorrect in handling this case

### DIFF
--- a/src/pymola/backends/casadi/generator.py
+++ b/src/pymola/backends/casadi/generator.py
@@ -546,8 +546,12 @@ class Generator(TreeListener):
         return s
 
     def get_derivative(self, s):
+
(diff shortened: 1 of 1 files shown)


vdwees:
@jackvreeken  there is more work to do to get the for-loops working, and I am planning to address them soon in another PR.

vdwees:
@jackvreeken is this ok to merge?

PR #50, opened 2018-02-08
Make tests also check alg_states for equivalence

jackvreeken:


jackvreeken:
It is just checking the length like the others, but I cannot think of a reason for not including it. I ran into one issue earlier yesterday where detecting it would have helped, which is why I made this PR. Is there anything you (@vdwees) can think of why we should _not_ include this check?

PR #52, opened 2018-02-15
Fix expansion of MX equations

jackvreeken:
According to Joris Gillis (CasADi co-author), ca.matrix_expand() is
broken and will not be fixed. Instead, we will then rely on regular
.expand() functionality to turn MX into SX, without the "graceful"
fallback that ca.matrix_expand() was supposed to offer. From initial
tests, it seems that most/all commonly used functionality is supported
by .expand().

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ sympy >= 0.7.6.1
 scipy >= 0.13.3
 numpy >= 1.8.2
 jinja2 >= 2.7.2
-casadi >= 3.3.0
+casadi >= 3.4.0
--- a/src/pymola/backends/casadi/model.py
+++ b/src/pymola/backends/casadi/model.py
@@ -68,6 +68,7 @@ class Model:
         self.initial_equations = []
         self.time = ca.MX.sym('time')
         self.delayed_states = []
+        self._expand_mx_func = lambda x: x
 
     def __str__(self):
         r = ""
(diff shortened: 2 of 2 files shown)


vdwees:
should expand be made default?

jackvreeken:
- [ ] Update CasADi depency in requirements.txt

1. We have to wait with merging until the version of CasADi on PyPI supports empty SXFunctions (i.e. no outputs). This is why the builds are currently failing.
2. Expanding should not be the default. MXFunctions are generally more compact than SXFunctions, and it's always easy to call .expand() to go from MX to SX. There is no way to go reverse from SX -> MX, so I'd rather keep it as is.
3. When serializable/picklable SXFunctions arrive in a PyPI version of CasADi—but this version does _not_ yet support pickling MXFunctions—I'll make `cache=True` imply `expand_mx=True` for the time being (and raise a warning/info message stating so). Expanding and caching will be decoupled again when support for pickling MXFunctions arrives in CasADi.
4. Related to (3); `codegen` and `cache` will become separate options. The former overrules the latter, but note that codegen can already work with both `expand_mx=True` and `expand_mx=False` (whereas `cache` will have to wait for picklable MXFunctions).

jackvreeken:
- [ ] Wait until CasADi 3.4.0 is released

jgoppert:
Can you fix the conflicts and rebase when you get a chance?

jackvreeken:
Apparently my other WIP was already merged (?), which was not supposed to happen before this one got merged.

vdwees:
for some reason #57 was merged before this one, so these changes are already incorporated. As I reviewed and approved both, I think we might as we as close this one.

jgoppert:
That was my fault, was trying to get as many PR's as I could merged last night and didn't know these were dependent on eachother.

PR #61, opened 2018-02-22
Add daebuilder in generator to allow simulation.

### LINKED ISSUE #58: Casadi based Simulation Support
Currently we provide the model equations, but don't have any mechanism to find the rhs for an ode, so it isn't straightforward to do simulation. Is there a simple way we can support this?

```python
ball = parser.parse("""
model Car
    input Real omega(start=0);
    input Real v;
    Real x1;
    Real x2;
    Real theta;
equation
    der(theta) = omega;
    der(x1) = v*cos(theta);
    der(x2) = v*sin(theta);
end Car;
""")
model = gen.generate(ball, 'Car')
eqs = model.equations

[MX((der(theta)-omega)),
 MX((der(x1)-(v*cos(theta)))),
 MX((der(x2)-(v*sin(theta))))]
```

We want to find the rhs:
```
rhs = [MX(-omega)),
 MX((-(v*cos(theta)))),
 MX((-(v*sin(theta))))]
```

So we can use it for simulation:
```
f = {'x':x,'t':t,'p': p,'ode':rhs}
Phi = ca.integrator("Phi","cvodes",f,{"tf":dt})
```

In the sympy backend, I was able to use sympy.solve(eqs_list, uknowns_list), and pass the state derivatives to the unknowns list, but I don't think there is a casadi equivalent.

jgoppert:
This  passes all the tests and it enables simulations for models that would be supported by daebuilder.

It basically intercepts any time you get a symbol and adds it to daebuilder. If you try to add something that isn't supported by dae builder (for instance a 2x3 matrix symbol) it will catch the exception and just print the error.

Closes #58 (note this requires casadi develop branch, so I can get the s and sdot terms out of add_s thanks to @jgillis).


jgoppert:
Demo here at the bottom: https://github.com/pymola/pymola/blob/sim/test/notebooks/Casadi.ipynb

jackvreeken:
1. Do you really want to use `DaeBuilder` instead of just calling `ca.integrator`? Because its sole author/maintainer does not deem it very mature, I'm afraid of relying on it. CasADi development can be fast paced, and also break things. If we use a higher level interface (DaeBuilder) to lower level functionality, it would be harder to make sure everything keeps working for us, or to leverage any newly added functionality.
2. This way we're supporting two different (sometimes conflicting) interfaces to the same data. People can ask for `model.parameters`, but also for `model.dae.p`. They may or may not give the same results, depending on simplification steps applied. I'd rather have one interface, so that people do not get confused (or even worse: mix them). It could be nice to replace ```self.model``` with ```self.dae``` and rely entirely on CasADi, but then we might have to shift a whole lot of functionality to CasADi instead. For example, we can no longer apply the simplifications, as we cannot modify the DaeBuilder object (or its attributes) directly (```self.model.p.append('ca.MX.sym(x)')``` does not do anything).
3. Conflict of naming. Currently it seems derivative symbols are automatically named "der_s", which is different from the current "der(s)". Would like to be able to name symbols myself.
	- The former is a valid identifier in Modelica, so seems undesirable.
	- 2D+ derivatives are still called "der(a)"
	- I am not sure how much of RTC-Tools depends on them being named "der(x)", but I'd rather not change this backwards compability.
4. If the user enables simplifications (e.g. vector/array expansion, parameter inlining, etc), it will mess with the equations of the DaeBuilder, but keep all variable lists as-is. I haven't tested, but it will probably break horribly.
5. DaeBuilder not supporting 2D seems limited.
6. With your changes we're checking on a symbol's prefixes twice now (and your new block misses a check for "constant"). An enum or something seems more appropriate if we would keep it.
7. DaeBuilder objects cannot be cached/saved. A reloaded model therefore lacks functionality that a freshly compiled model does offer.
8. The current `self.model` does not directly expose CasADi, but uses `Variable` objects to enable e.g. aliasing which would otherwise not be possible. By using DaeBuilder or the underlying CasADi graph directly, one will lose this functionality, which seems like a waste.

Some points being more important than others, I'd consider the CasADi backend currently more stable and feature-rich than the DaeBuilder. I'd therefore rather not make this switch, and instead just construct the necessary information to `ca.integrator` ourselves for now. As I showed in https://github.com/pymola/pymola/issues/58#issuecomment-367313680 this should not be that difficult. The only thing left then is to figure out the "make_explicit" thing, or expose that functionality in CasADi in a different way such that it can be used without making a DaeBuilder object.

If Joel ends up reworking DaeBuilder as part of his current/future endeavours, it would make sense to reconsider and e.g. turn ```self.model``` into a DaeBuilder object. We would have to be able to modify this object and its variables/equations though, although I really do not know how we would ever be able handle aliases with it.

jgoppert:
@jackvreeken OK, I'll start looking at how to add a make explicit function to the model class. I think now that I've been digging around in here for awhile it should be possible.

jackvreeken:
Yes, you definitely are getting familiar with the CasADi backend (and CasADi itself) at an astounding pace. I've been working with CasADi a bit more more intensively for a few months now, and I still keep getting lost...

PR #62, opened 2018-02-22
More aggressive alias checking

### LINKED ISSUE #55: Fix alias detection of SimplifyLoop.mo
Test `test_simplify_reduce_affine_expression_loop` uses alias detection simplification, but `y[1]` is not detected as an alias of `x`. Probably because of the `vector_expand` resulting in horribly convoluted expressions, which we may need to work around.

jackvreeken:
If we cannot detect an alias based on the MX symbols/expression, it may
be because of nested map/if_else/etc. expressions that obscure it. To
expose remaining possible aliases, we expand the equation to SX.

Note that we still do fast checks first, e.g. on the MX graph. The
slowest operation (ca.substitute) is only used as a last resort. This
way the slowdown to the alias detection is is minimal (1.0 - 1.5x), even
for very large models with 100,000+ alias equations.

These extra checks typically result in 1-3% more aliases being detected.

The 'test_simplify_reduce_affine_expression_loop' now finds an
additional alias (x = y[1]). To be able to reliably keep testing the
'reduce affine' portion, the size of 'y' has been increased from 2 to 3.

Closes #55

### DIFF
--- a/src/pymoca/backends/casadi/model.py
+++ b/src/pymoca/backends/casadi/model.py
@@ -124,7 +124,195 @@ class Model:
         durations = ca.substitute([ca.MX(argument.duration) for argument in delay_arguments], symbols, values)
         return [DelayArgument(expr, duration) for expr, duration in zip(exprs, durations)]
 
+
+    @staticmethod
+    def _expand_simplify_mx(equations):
+        # Sometimes MX expressions can end up horribly nested and complicated,
+        # which makes further simplifications like alias detection difficult.
+        # We can simplify the equations by expanding to SX, and then
+        # rebuilding them with MX symbols again.
+
+        if not equations:
+            return []
+
+        symbols_mx = ca.symvar(ca.veccat(*equations))
+        assert all(x.shape == (1, 1) for x in symbols_mx), "Vector/Matrix SX symbols cannot be mapped"
+
+        f_mx = ca.Function('tmp_mx', symbols_mx, equations).expand()
+        symbols_sx = [ca.SX.sym(x.name(), *x.shape) for x in symbols_mx]
+        sx_equations = f_mx.call(symbols_sx)
+
+        sx_to_mx_map = {s: m for s, m in zip(symbols_sx, symbols_mx)}
+
+        def _sx_to_mx(sx_expr):
+            if not sx_expr.is_scalar():
+                rows = []
+                for i in range(sx_expr.size1()):
+                    cols = []
+                    for j in range(sx_expr.size2()):
+                        cols.append(_sx_to_mx(sx_expr[i, j]))
+                    rows.append(cols)
+                return ca.vertcat(*[ca.horzcat(*row) for row in rows])
+            elif sx_expr.op() == ca.OP_PARAMETER:
+                assert sx_expr in sx_to_mx_map.keys()
+                return sx_to_mx_map[sx_expr]
+            elif sx_expr.op() == ca.OP_CONST:
+                return ca.MX(ca.DM(sx_expr))
+            elif sx_expr.n_dep() == 1:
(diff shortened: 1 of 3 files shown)


jackvreeken:
- [ ] Wait for CasADi 3.4.0 to be released, as we need support for `__hash__` of SX symbols.

vdwees:
@jackvreeken This is an improvement for sure, but it could go a lot farther- the logic addresses more than just aliases. for example, in Rijnland,
```modelica
MX(switch(0, E_BG_Halfweg.Q, E_BG_Halfweg.HQ.M, 0, E_BG_Halfweg.HQ.C){0})
```
becomes
```modelica
SX(E_BG_Halfweg.HQ.M)
```

This is not an alias, but rather a alg_state that ought to have been caught here:


```python
if options.get('eliminate_constant_assignments', False):
    logger.info("Elimating constant variable assignments")

    alg_states = OrderedDict([(s.symbol.name(), s) for s in self.alg_states])

    reduced_equations = []
    for eq in self.equations:
        if eq.is_symbolic() and eq.name() in alg_states:
            constant = alg_states.pop(eq.name())
            constant.value = 0.0

            self.constants.append(constant)
```

Furthermore, the equation
```modelica
MX(vertsplit((vertcat(Aarkanaal_1._cross_section[0], Aarkanaal_1._cross_section[1])-(all_29.0701(2x1)*(vertcat(Aarkanaal_1.H[0], Aarkanaal_1.H[1])-all_-3.1178(2x1))))){1})
```
Is overly complicated as an MX expression. Upon expanding, it becomes
```modelica
SX((Aarkanaal_1._cross_section[1]-(29.0701*(Aarkanaal_1.H[1]--3.1178))))
```
This is much nicer- why not save that equation?

So while the logic of this MR is ok, I think we need it in a lot of different places. Or, alternatively, right away, before we run through all the simplify() algorithms.

jgoppert:
Ready to merge this?

jackvreeken:
Not yet.

jgoppert:
@jackvreeken should we wait on releasing 0.4 for this?

jackvreeken:
No, this is not crucial in any way. It's more of a 'nice to have', which could lead to some performance improvements for us.

jackvreeken:
A check that gets even more (than just the SX):

```python
x = ca.SX.sym('x')
y = ca.SX.sym('y')
z = ca.SX.sym('z')

f = x * (y - 2 * z)
g = x * (y - 2 * z)

# Not good enough yet
f - g  # SX(@1=2, ((x*(y-(@1*z)))-(x*(y-(@1*z)))))
(f - g).is_zero()  # False
# Expanding also does not work
f_exp = ca.Function('tmp', [x, y, z], [f - g]).expand()
f_exp(x, y, z)  # SX(@1=2, ((x*(y-(@1*z)))-(x*(y-(@1*z)))))

# What does work is using the pickling/serializing on SXFunctions
f_f = ca.Function('tmp', [x, y, z], [f])
f_g = ca.Function('tmp', [x, y, z], [g])

f_f.serialize() == f_g.serialize()  # True
```

jackvreeken:
New implementation:
- When expand_vectors and expand_mx are both True, we do that combination first. This results in relatively pretty MX expressions, which addresses @vdwees comment about keeping the prettier equations.
- When expand_mx is False, we cannot rewrite equations (as that would inline function calls, etc). We can still however temporarily expand each equation to see if it describes an alias. If it doesn't, we don't touch it. If it does, we remove it.

Note that the more aggressive alias check does _not_ work for vector expressions due to the mapping of SX symbols to MX symbols (and a vector SX symbol does not have a valid `__hash__`). This generally does not matter, as it's the very expansion of vectors that results in the ugly equations in the first place.

PR #63, opened 2018-02-23
pycharm inspection fixes

jgoppert:


src/pymola/tree.py
```diff
@@ -189,7 +189,7 @@ def walk(self, listener: TreeListener, tree: ast.Node) -> None:
         if hasattr(listener, 'enter' + name):
             getattr(listener, 'enter' + name)(tree)
         for child_name in tree.__dict__.keys():
-            if isinstance(tree, ast.Class) and child_name in ('parent') or \
+            if isinstance(tree, ast.Class) and child_name in 'parent' or \
```
jackvreeken: Instead of `in 'parent'` it should probably be `== 'parent'`. The reason for the `in ...` is because this was copied from the check in `ast.py` iirc, where it checks on `('parent', 'root')`.

src/pymola/parser.py
```diff
@@ -23,7 +23,7 @@
 #  - Named function arguments (note that either all have to be named, or none)
 #  - Make sure slice indices (eventually) evaluate to integers
 
-class ModelicaFile:
+class ModelicaFile(object):
```
jackvreeken: We don't really care about new/old style classes (a Python 2 thing) anymore do we?

src/pymola/backends/casadi/model.py
```diff
@@ -52,7 +54,8 @@ def from_dict(cls, d):
 
 
 # noinspection PyUnresolvedReferences
-class Model:
+class Model(object):
```
jackvreeken: Why inheriting from object? Python 2 vs 3 thing?

src/pymola/tree.py
```diff
@@ -739,7 +738,8 @@ def exitSymbol(self, tree: ast.Symbol):
     #def exitClassModificationArgument(self, tree: ast.ClassModificationArgument):
     #     assert isinstance(tree.value, ast.ElementModification), "Found unhandled redeclaration in instance tree."
 
-    def exitInstanceClass(self, tree: ast.InstanceClass):
+    @staticmethod
+    def exitInstanceClass(tree: ast.InstanceClass):
```
jackvreeken: This does not look right. It should have the same signature as all the other methods. If PyCharm suggests this, there's probably something wrong somewhere else?

jackvreeken:
We can consider going one step further by fixing a bunch of other formatting/naming/etc stuff as well, such that we can enable a flake8 test. I ran it at one point, but it returned a whole lot of errors. As far as line-length is concerned, I would like 120 (the default on GitHub and GitLab).

jgoppert:
Yeah, I was going to do some formatting next, but we need to make sure all of the branches are in first so we don't end up with a merge nightmare. I would vote we keep the current default line length for pycharm as that is  how most of it is formatted currently and it usually works better for me when I am doing split screen development, but not a huge deal.

jackvreeken:
That's great then, as PyCharm's default line length is also 120 characters :smile:  I agree that formatting should go in after the WIP PRs (that are waiting on CasADi 3.4.0) are merged.

PR #64, opened 2018-02-25
Add rudimentary support for higher order arrays

### LINKED ISSUE #53: Add support for 3D+ arrays in CasADi backend
Initially we probably only need support for 3D+ parameters, e.g.:

```Modelica
      parameter Real working_area[4, 2, 3] = {{{ -5.326999,  54.050758,  0.000000},
                       { -1.0,        0.0,       0.0}},
                      {{  0.000426,  -0.001241,  2.564056},
                       { -1.0,        0.0,       0.0}},
                      {{  2.577975,  -5.203480,  0.000000},
                       { -1.0,        0.0,       0.0}},
                      {{ 13.219650,  -3.097600, -7.551339},
                       { -1.0,        0.0,       0.0}}},
```

If needed, we can make a MTensor class that wraps a (flat) MX with appropriate dimensions.

jackvreeken:
This only adds support for higher ordered parameters and values, but
does not allow users to use these symbols in expressions.

Closes #53 

- [x] Make MTensor internal
- [x] Check that MTensor is only used in conjunction with expand_vectors.


jackvreeken:
I am not entirely happy with the way CasADi MX behaves. It's modeled very much after Matlab, but it would be much more convenient for us if it were more like NumPy instead. Scalars, and 1-D vectors always having two dimensions (`x.shape` where `x` is a `ca.MX`) is weird, and not supporting higher dimensions is also.

As far as ambiguous `Real[3]` times `Real[3]` is concerned, Modelica handles that as follows (see  http://book.xogeny.com/behavior/arrays/functions/ ):
1. Normal multiplication operator behaves as dot procuct.
2. To get 3x3 matrix as result, use `outerProduct(a, b)` function.

src/pymoca/backends/casadi/api.py
```diff
@@ -300,6 +300,14 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
                             and ca.depends_on(metadata[key][i, j], parameter_vector)):
                             setattr(variable, tmp, metadata[key][i, j])
 
+        # Try to coerce parameters into their Python types
+        for p in model.parameters:
+            for attr in CASADI_ATTRIBUTES:
+                v = getattr(p, attr)
+                v_mx = ca.MX(v)
+                if v_mx.is_constant() and v_mx.is_regular():
+                    setattr(p, attr, p.python_type(v))
```
jbaayen: Useful, but does this belong in this PR?

src/pymoca/backends/casadi/mtensor.py
```diff
@@ -0,0 +1,29 @@
...
+    @property
+    def shape(self):
+        return self._shape
+
+    def __getattr__(self, attr):
+        return getattr(self._mx, attr)
+
+    def __getitem__(self, k):
+        assert len(k) == len(self._shape)
+        return self._mx[np.ravel_multi_index(tuple(k,), self._shape)]
```
jbaayen: Neat!

src/pymoca/backends/casadi/mtensor.py
```diff
@@ -0,0 +1,29 @@
+import casadi as ca
+
+import numpy as np
+
+
+def new_mx(name, *shape):
```
jbaayen: Check that len(shape) <= 2

jbaayen: Does this method belong in this file, mtensor.py?

Have you considered setting up a general wrapper class that makes the decision whether to create a 'classical' MX or a flattened one internally? The class could be set up as a decorator for the underlying MX. We could call it 'TX', for example.

src/pymoca/backends/casadi/model.py
```diff
@@ -108,12 +109,21 @@ def simplify(self, options):
 
             simple_parameters, symbols, values = [], [], []
             for p in self.parameters:
-                value = ca.MX(p.value)
-                if value.is_constant():
+                if isinstance(p.value, list):
```
jbaayen: Why introduce this extra check?

```
In [6]: ca.MX([1,2,3]).is_constant()
Out[6]: True
```

src/pymoca/backends/casadi/model.py
```diff
@@ -332,37 +342,29 @@ def simplify(self, options):
...
+                    if old_var.symbol._modelica_shape != (1,):
+                        expanded_symbols = []
+                        for ind in np.ndindex(old_var.symbol._modelica_shape):
+                            component_symbol = ca.MX.sym('{}[{}]'.format(old_var.symbol.name(), ",".join(str(i+1) for i in ind)))
+                            component_var = Variable(component_symbol, old_var.python_type)
+                            for attribute in CASADI_ATTRIBUTES:
+                                # Can't convert 3D arrays to MX, so we convert to nparray instead
+                                value = getattr(old_var, attribute)
+                                if not np.isscalar(value):
+                                    value = np.array(value)
```
jbaayen: Does this work when value contains symbolic MX?

src/pymoca/backends/casadi/model.py
```diff
@@ -332,37 +342,29 @@ def simplify(self, options):
...
-                        values.append(ca.vertcat(*[ca.horzcat(*[v.symbol for v in row]) for row in rows]))
-                        new_vars.extend(itertools.chain.from_iterable(rows))
+
+                                if np.prod(value.shape) == 1:
+                                    setattr(component_var, attribute, value)
+                                else:
+                                    setattr(component_var, attribute, value[ind])
+                            expanded_symbols.append(component_var)
+
+                        s = old_var.symbol._mx if isinstance(old_var.symbol, MTensor) else old_var.symbol
```
jbaayen: This would be much nicer if we had this single "TX" wrapper class (see below)

src/pymoca/backends/casadi/api.py
jackvreeken: I just turned this PR into stuff to get JSM to work. And there we need integer parameters to be integers, so this change is needed.

src/pymoca/backends/casadi/model.py
jackvreeken: Doing that on higher order arrays will fail. E.g. `ca.MX(np.array([[1,2,3],[4,5,6]]))`

src/pymoca/backends/casadi/mtensor.py
jackvreeken: Already covered by CasADi itself.

jackvreeken: I tried making it pretty with a wrapper class, but kept running into weird SWIG related stuff (overriding `__new__` did not work). So instead I opted to do it this way instead.

src/pymoca/backends/casadi/model.py
jackvreeken: Probably not, but using parameters inside array values is failing for different reasons now anyway.

src/pymoca/backends/casadi/api.py
jbaayen: Hmm, but can we validly assume integer parameters to evaluate to integers in the CasADi/Python world?

I have my doubts. Consider of this snippet:

```
parameter Integer n = 2;
parameter Integer m = 2 * n +1;
```

clearly `m` is also an integer, but it will be a casadi expression, thereby evaluating to an (integer valued) float.

jbaayen: ... so I wonder if we shouldn't be casting to integers in the JSM model code instead.

src/pymoca/backends/casadi/mtensor.py
jbaayen: Instead of subclassing, how about a simple wrapper class with an `mx` property?

src/pymoca/backends/casadi/model.py
jbaayen: Gotcha.

src/pymoca/backends/casadi/model.py
jbaayen: OK, but in any we shouldn't introduce additional array value issues with new code.

src/pymoca/backends/casadi/api.py
jackvreeken: Eh, maybe. What this commit does (like with the variable metadata commit) is to make sure the behavior of the _uncached_ model matches that one of the _cached_ one. That's why it's  called "coerce cached ... into their python types".

I'm not sure what to do about the `m` case. When inlining of parameters in Pymoca it should imo end up as `int`. In the general case where it remains an expression until the very end (i.e in the DAE), I do think we want a `ca.floor` around it to make sure it actually is an integer. That said, I'm not sure what the following does or should do.

```
parameter Integer a = 1;
parameter Integer b = a / 2;
```
Does that make `b` zero, or should it raise an error? Either way, I would say this (and the `m` stuff) is a new issue. Again, this commit just makes sure that the cached version matches the uncached/raw version.

src/pymoca/backends/casadi/model.py
jackvreeken: Possibly, but the only thing I can do now is raise an `NotImplementedError`. I could work the way it is, but with other things (before it would even get here) broken, it is hard to test and be sure.

PR #70, opened 2018-03-07
Cleanup setup.py

jackvreeken:


jackvreeken:
@jgoppert : Do you know where `try-except` thing came from, and the `sys.path` modification? It didn't seem to do anything (anymore), but maybe you added is as a workaround for something at one point?

jgoppert:
I went and looked at the blame, and it is 4 year old boilerplate I believe. I think we can delete it now that we are just calling setup directly without the dictionary.

jackvreeken:
OK. I just squashed everything in one commit now (as it's all just removing obsolete code/cruft from one particular file). Will merge when tests succeed.

PR #71, opened 2018-03-07
WIP: Add flake8 formatting test

jackvreeken:
- [ ] Fix "failing" lines
- [x] Line limit of 119 or 120? --> 99 to comply with PEP8
- [x] Do we also want to use `flake8-quotes`? --> No.
- [ ] `flake8-import-order` vs `flake8-isort`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,4 @@ Untitled.ipynb
 .pytest_cache/
 jit_tmp.c
 tmp*.o
+*.orig
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,11 @@ matrix:
         - python3 -m venv venv
         - source venv/bin/activate
       env: TOXENV=py36
-    # TODO: Meta for flake8
+
+    # Meta
+    - os: linux
+      python: 3.6
+      env: TOXENV=flake8
 
 before_install:
   - pip install coveralls
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,3 +12,39 @@ versionfile_source = src/pymoca/_version.py
 versionfile_build = pymoca/_version.py
 tag_prefix =
 parentdir_prefix = pymoca-
+
+
+[flake8]
+#ignore = D203
+max-line-length = 99
+exclude =
+	# misc
+	.git,
+	.tox,
+	docs/source/conf.py,
+	# generated files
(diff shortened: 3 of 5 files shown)


jgoppert:
After running flake8 on the project and seeing all of our lines over 79, I get why you are pushing for 119 or 120, but I think it should be a goal to get to 79 eventually to meet PEP8, but definitely not worth the effort to convert everything right now.

jackvreeken:
In practice I find 80 to be very short, and 100-120 to be quite okay still. With the de facto standard line length in GitLab, GitHub, and PyCharm being 120, I figured I was not the only one that felt that way. And PEP8 also states that longer line lengths are allowed (not sure about 120, but 100 is OK), but that docstrings should still be wrapped at 72 characters.

I personally don't feel particularly strong about adhering to PEP8 on this issue, but I do like enforcing a consistent coding style. So if you would like to adhere to PEP8 on the line length, I would suggest going for 99 instead of 79.

Any opinion on the quotes (single vs. double)? PEP8 allows both, but some developers strongly prefer one over the other (e.g. double to mimic C++/Java).

jgoppert:
I agree we should be consistent on the quoting. Let's go with double quotes.
99 sounds good for line length.

vdwees:
I like single quotes, as they are prettier and don't require a shift key... but double is ok too.

jackvreeken:
Added quotes checking. As I said, I personally tend to use both (and don't mind doing so), but setting a convention is fine by me as well.

Note that backslashes in strings to escape quotes is very much discouraged by PEP8, so the following two options are always OK according to `flake8-quotes`:
```Python
s = 'double "quotes" wrapped in singles are ignored'
s = "single 'quotes' wrapped in doubles are ignored"
```

jgoppert:
I'm fine with ignoring or single quotes too. I also find myself using both. I agree single are prettier and easier to type.

jackvreeken:
Then I'll just remove that check. If at least two of us are using both double and single quotes without really caring/noticiing, it does not make much sense to enforce one over the other (especially because PEP8 does not say anything about it either).

jgoppert:
I set this to explicitly ignore the files with known issues and fixed setup.py. So we can merge this to make sure new code conforms.

PR #97, opened 2018-03-13
Update roadmap.

jgoppert:


jgoppert:
One thing we might want to add is FMU/FMI support eventually, but this might require us writing a c-backend, put perhaps we could use casadi to write the backend for us. This would allow users to directly use our models in matlab simulink etc, which many would find useful.

jgoppert:
The FMU idea actually wasn't too bad, I did a prototype here: https://github.com/jgoppert/pymoca_fmu. Since I have to redo the connection to sundials to support root solving for my problem, might as well do it all in c++. https://github.com/jgoppert/pymoca_fmu  Maybe eventually it will make sense to make it part of the project, if not, can just keep it separate.

Issue #103, opened 2018-03-15
CasADi backend needs to support code generation starting from an AST


jbaayen:
Right now, `transfer_model()` takes a filename as input.

We need a variant of this function that takes an AST as input, to allow programmatically created ASTs to be compiled to CasADi and code-generated.

Issue #102, opened 2018-03-14
Validate AST and Specify API


jgoppert:
I think it would be good to start the AST API discussion. I'm finding myself expanding array expressions for the xml backend which I know has already been done for the casadi backend. It would be nice if this was handeled on the AST side.

If we look at the ModelicaXML tree descriptions for a guide. I think it would make sense to be able to ask  the pymoca API for the tree at any stage 1, 2, 3 or 4. This would let us share the array expansion work between backends, which is my current motivation.

> 1. Modelica source code. This format is an abstract representation of Modelica code and should provide a one to one mapping between Modelica source code and a corresponding XML representation.
> 2. Instantiated hierarchical models. In this model representation, Modelica modifiers have taken effect, and the hierarchical structure is preserved.
> 3. Flattened Modelica models. This model representation is flat in the sense that it does not contain hierarchical model elements. It may, however, contain array declarations.
> 4. Flattened transformed models. In this representation, array variables and equations have been scalarized, alias variables have been eliminated and DAE index has been reduced, if needed.

Maybe something like this?
```python
# step 1
ast_code = parse(pymola)
generate_xml_code(ast_code).validate()

# step 2
ast_inst = instantiate(ast_code)
generate_xml_inst(ast_int).validate()

# step 3
ast_flat = flatten(ast_inst)
generate_xml_flat(ast_flat).validate()

# step 4
ast_reduced = reduce(ast_flat)
generate_xml_reduced(ast_reduced).validate()

# send to backends
generate_casadi(ast_flat... or ast_reduced, whatever is easiest)
```

jbaayen:
For AST manipulation, I'd like to see equivalents for at least the following:

* `ast.Class.{add,remove}_class(cls: ast.Class)`
* `ast.Class.{add,remove}_symbol(sym: ast.Symbol)`
* `ast.Class.{add,remove}_equation(equation: ast.Equation)`

jackvreeken:
These steps are currently implemented in two places: The flattener/`tree.py` handles (2) instantiation and (3) flattening, and the CasADi backend/`model.py` handles the array expansion. Do I understand you correctly that you would want to move step 4 to be _before_ handing it off to the backend? I think it is going to be tough to do alias detection/for loops/etc without relying on a symbolic framework like CasADi.

[...]

PR #104, opened 2018-03-16
AST manipulation API

jbaayen:
Addresses #103 and a part of #102.

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -3,6 +3,7 @@ from typing import Dict
 import distutils.ccompiler
 import casadi as ca
 import numpy as np
+import copy
 import sys
 import os
 import fnmatch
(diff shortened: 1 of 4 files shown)


src/pymoca/backends/casadi/api.py
```diff
@@ -145,9 +145,20 @@ def _codegen_model(model_folder: str, f: ca.Function, library_name: str):
...
+
+    :param model_folder: Folder where the precompiled CasADi model will be stored.
+    :param model_name: Name of the model.
+    :param model: Model instance.
+    :param compiler_options: Dictionary of compiler options.
+    """
+
     # Compilation takes precedence over caching, and disables it
+    cache = compiler_options.get('cache', False)
+    codegen = compiler_options.get('codegen', False)
```
jackvreeken: Saving a model only makes sense if one of them is True. This was not necessary in `transfer_model` (as it would just _not_ save it), but should probably be part of the `save_model` if we make it public (and remove the default arguments).

test/ast_test.py
```diff
@@ -0,0 +1,36 @@
...
+        s = ast.Symbol(name='TestSymbol', type=ast.ComponentRef.from_tuple('Real'))
+        self.ast.add_symbol(s)
+
+        self.assertIn(s.name, self.ast.symbols)
+
+        self.ast.remove_symbol(s)
+        self.assertNotIn(s.name, self.ast.symbols)
+
+    def test_add_equation(self):
+        e = ast.Equation(left=ast.ComponentRef.from_tuple('a'), right=ast.ComponentRef.from_tuple('b'))
```
jackvreeken: Flake8 test already mentioned this, but this line is too long.

src/pymoca/backends/casadi/api.py
```diff
@@ -145,18 +146,22 @@ def _codegen_model(model_folder: str, f: ca.Function, library_name: str):
...
+    :param model: Model instance.
+    :param compiler_options: Dictionary of compiler options.
+    """
 
     objects = {'dae_residual': None, 'initial_residual': None, 'variable_metadata': None}
     for o in objects.keys():
         f = getattr(model, o + '_function')
 
-        if codegen:
+        if compiler_options['codegen']:
```
jackvreeken: Shouldn't this be `compiler_options.get('codegen', False)`?

PR #110, opened 2018-05-02
Fix caching of variable attributes

### LINKED ISSUE #109: casadi load_model() api replaces parameter values in expressions
When loading variable metadata from a cached model, the ```load_model()``` method does three calls of the ```variable_metadata_function()```. In the process, symbolic expressions that contain parameters with known values are resolved into a numeric value prematurely, so that parameter values cannot be overwritten later. Effectively, cached models behave as though the ```replace_parameter_values``` compiler option is true.

jackvreeken:
- [x] Add (previously failing) test cases for both commits

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -266,39 +266,18 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
         model.delay_states = db['delay_states']
 
         # Evaluate variable metadata:
-        # We do this in three passes, so that we have constant attributes available through the API,
-        # and non-constant expressions as Function calls.
-
-        # 1.  Extract independent parameter values
-        metadata = dict(zip(variables_with_metadata, model.variable_metadata_function(ca.veccat(*[np.nan for v in model.parameters]))))
-        for key in variables_with_metadata:
-            for i, d in enumerate(db[key]):
-                variable = variable_dict[d['name']]
-                for j, tmp in enumerate(CASADI_ATTRIBUTES):
-                    setattr(variable, tmp, metadata[key][i, j])
-
-        # 2.  Plug independent values back into metadata function, to obtain values (such as bounds, or starting values)
-        # dependent upon independent parameter values.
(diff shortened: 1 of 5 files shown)


src/pymoca/backends/casadi/api.py
```diff
@@ -257,39 +257,18 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
...
                 variable = variable_dict[d['name']]
                 for j, tmp in enumerate(CASADI_ATTRIBUTES):
-                    if not getattr(variable, tmp).is_regular():
-                        if (not isinstance(metadata[key][i, j], ca.DM)
-                            and ca.depends_on(metadata[key][i, j], parameter_vector)):
-                            setattr(variable, tmp, metadata[key][i, j])
+                    if ca.depends_on(metadata[key][i, j], parameter_vector):
+                        setattr(variable, tmp, metadata[key][i, j])
+                    else:
+                        setattr(variable, tmp, independent_metadata[key][i, j])
```
jbaayen: So what happens now is that parameters of the form

`parameter Real a = 1.0;`

get their numerical value exposed.

OTOH parameters like

`parameter Real b = a;`

do not. The original intent was to produce the value `1.0` when accessing `b` as well.

So this is an API change which will have ramifications for RTC-Tools as well as some applications. For instance, now, any calculations in `pre()` using parameter values may fail if there is an in-between assignment anywhere.

jackvreeken: A cached model should have exactly the same attributes as a freshly parsed model. This PR is addressing that discrepancy. If the behavior of a freshly parsed model should be different, then we should discuss and address that separately.

jbaayen: I thought the behaviour was the same at some point in time, but perhaps I am mistaken.

I agree about consistency but bear in mind that this change may cause breakage in RTC models which expect pymoca to behave like JModelica in this regard.

jackvreeken: There were no checks in place for the correctness of the `variable_metadata_function`, and also none for the additional steps in `load_model()`. It's therefore hard to say whether they were ever equal.

PR #113, opened 2018-05-08
Support delay() in for loops

jackvreeken:
~~It's a very hacky implementation, specifically to get Channel Flow to work. It still does not properly run however, and I have no clue why. I'll put this on hold for now, until I find some more time/energy to work on this.~~

- [ ] Wait for #118  to get merged, as this PR is on top of that one.

### DIFF
--- a/src/pymoca/backends/casadi/generator.py
+++ b/src/pymoca/backends/casadi/generator.py
@@ -52,7 +52,7 @@ class ForLoop:
         self.values = np.arange(start, stop + step, step, dtype=np.int)
         self.index_variable = ca.MX.sym(i.name)
         self.name = i.name
-        self.indexed_symbols = {}
+        self.indexed_symbols = OrderedDict()
 
     def register_indexed_symbol(self, e, index_function, transpose, tree, index_expr=None):
         if isinstance(index_expr, ca.MX) and index_expr is not self.index_variable:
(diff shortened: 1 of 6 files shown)


jbaayen:
Works fine for me with a simple example.

src/pymoca/backends/casadi/generator.py
```diff
@@ -248,14 +249,24 @@ def exitExpression(self, tree):
                     entries.append(sym)
             src = ca.vertcat(*entries)
         elif op == 'delay' and n_operands == 2:
+            # Delay can accept an arbitrary expression, whereas we only seem to support states?
```
jbaayen: See #115

src/pymoca/backends/casadi/generator.py
```diff
@@ -355,7 +366,18 @@ def exitForEquation(self, tree):
             indexed_symbols_full = []
             for k in indexed_symbols:
                 s = f.indexed_symbols[k]
-                orig_symbol = self.nodes[self.current_class][s.tree.name]
+                if k.name() in [x.name for x in self.model.delayed_states]:
```
jbaayen: Use generator here?

src/pymoca/backends/casadi/model.py
```diff
@@ -50,9 +50,11 @@ def from_dict(cls, d):
...
         return variable
 
-
-DelayedState = namedtuple('DelayedState', ['name', 'origin', 'delay'])
-
+class DelayedState:
+    def __init__(self, name, origin, delay):
+        self.name = name
+        self.origin = origin
+        self.delay = delay
```
jbaayen: Why this?

src/pymoca/tree.py
```diff
@@ -10,6 +10,7 @@
 import sys
 from collections import OrderedDict
 from typing import Union
+import os
```
jbaayen: No need for this I suppose

src/pymoca/backends/casadi/model.py
```diff
@@ -361,9 +363,19 @@ def simplify(self, options):
                             rows.append(cols)
                         symbols.append(old_var.symbol)
                         values.append(ca.vertcat(*[ca.horzcat(*[v.symbol for v in row]) for row in rows]))
-                        new_vars.extend(itertools.chain.from_iterable(rows))
+                        new_vars_extra = list(itertools.chain.from_iterable(rows))
+                        if l == 'inputs':
```
jbaayen: I don't understand. We can also delay regular states or even algebraic states.

src/pymoca/backends/casadi/generator.py
```diff
@@ -355,7 +366,18 @@ def exitForEquation(self, tree):
...
+                    # We are missing a similarly shaped delayed symbol. Replace/make it.
```
jbaayen: We should probably change things such that the symbol in `self.model.delayed_states` has the right dimension for the for loop.

src/pymoca/backends/casadi/model.py
jackvreeken: Have to overwrite field. Not possible with tuple.

src/pymoca/backends/casadi/model.py
jackvreeken: Yes, but the test case was with inputs

jackvreeken:
Picked up further in #118

PR #118, opened 2018-05-17
Reconstruct delay_arguments in load_model(), and handle delay arguments

jbaayen:
in simplify().

- [x] Test the substitute for delay variables in simplify()
- [x] Test the writing/reading delay from cache

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -13,7 +13,7 @@ import contextlib
 
 from pymoca import parser, tree, ast, __version__
 from . import generator
-from .model import CASADI_ATTRIBUTES, Model, Variable
+from .model import CASADI_ATTRIBUTES, Model, Variable, DelayArgument
 
 logger = logging.getLogger("pymoca")
 
@@ -30,6 +30,7 @@ class CachedModel(Model):
         self.parameters = []
         self.time = ca.MX.sym('time')
         self.delay_states = []
+        self.delay_arguments = []
 
         self._dae_residual_function = None
         self._initial_residual_function = None
@@ -72,10 +73,6 @@ class CachedModel(Model):
     @property
     def initial_equations(self):
         raise NotImplementedError("Cannot access individual equations on cached model.  Use residual function instead.")
-
-    @property
-    def delay_arguments(self):
-        raise NotImplementedError("Cannot access delay arguments on cached model.  Use delay arguments function instead.")
 
     def simplify(self, options):
         raise NotImplementedError("Cannot simplify cached model")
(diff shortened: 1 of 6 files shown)


jackvreeken:
Looking at the changes (and coveralls failing), it makes sense to make new test cases, or adjust existing ones to

1. test the substitute for delay variables in simplify() 
2. test the writing/reading delay related things from cache, and verify that cached == non-cached (also for the delayed states).

vdwees:
I would prefer if the _substitute_delay_arguments() returned delay_arguments:

```python
self.delay_arguments = _substitute_delay_arguments(symbols, values)
```
It avoids relying on side-effects and follows the `ca.substitute()` pattern.

jbaayen:
If you do that, then I'd suggest to make it fully functional by also passing in `self.delay_arguments`.

vdwees:
True. Good catch.

vdwees:
something fishy is happening when loading cached models

Uncached:
```
DelayArgument(expr=MX(Q_in), duration=MX(delay_as_input))
```

Cached:
```
[DelayArgument(expr=MX(@1=0x0, delay_arguments(time, @1, @1, @1, vertcat(_pymoca_delay_0, Q_in, delay_as_input), @1, @1){0}[0]), duration=DM(nan))]
```

jackvreeken:
The delay duration is assumed to be a function of parameters. I'm not very happy about the inlining parameter values here again (like it used to be with variable metadata), please get rid of that:

```Modelica
        for i in range(len(args)):
            args[i] = ca.repmat(np.nan, *args[i].size())
        args[6] = ca.veccat(*[v.value if v.value.is_regular() else v.symbol for v in model.parameters]) // No, don't do this
        independent_delay_arguments_raw = model.delay_arguments_function(*args)
```

Otherwise, what is there now is probably correct, which is that `delayTime` should be a function of parameters if `delayMax` is not specified [0]. In our case however, it also needs to be a function of `inputs` that have `fixed=True` set. Of course, if `delayMax` is set, it could be an arbitrary (???) expression. But as far as I know, we do not support delay() with 3 operators, so we cannot handle the case with `delayMax`: 

```Python
        for i in range(len(args)):
            args[i] = ca.repmat(np.nan, *args[i].size())
        args[4] = ca.veccat(*[v.symbol if v.fixed else np.nan for v in model.inputs])
        args[6] = ca.veccat(*[v.symbol for v in model.parameters])
        independent_delay_arguments_raw = model.delay_arguments_function(*args)
```

One request I have though is to check somewhere in `model.py` or `generator.py` that the duration is indeed only an expression of fixed inputs and parameters (assuming that delayMax is not supported at all, or not set).

source:
[0] https://build.openmodelica.org/Documentation/ModelicaReference.Operators.%27delay()%27.html

vdwees:
Ok, constant inputs are working now. Next stop, for loops.

src/pymoca/backends/casadi/api.py
```diff
@@ -301,6 +301,13 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
 
                 if ca.depends_on(ca.MX(delay_durations_raw[i]), all_symbols):
                     dur = delay_durations_raw[i]
+                    false_deps = ca.vertcat(*[
+                        var for var in ca.symvar(dur) if not ca.depends_on(dur, var)])
```
jackvreeken: That's an interesting way of doing it. This would hopefully also catch parameters and fixed inputs that aren't used in the durations. I do hope that `ca.depends_on` is accurate though, and stays accurate... I would've initially guessed that `depends_on` and e.g. `sparsity_jac()` would share code, but due to their different results that is maybe not the case.

vdwees: Yes, it was important to also catch the parameters and fixed inputs.
I agree, relying on depends_on() is a small gamble, but it is working well so far.

Issue #127, opened 2018-05-25
Nested parameters not looked up correctly in CasADi backend


jackvreeken:
Currently the CasADi backend just assumes that a nested parameters `x` inside an object (e.g. `a[:].x`) are all one and the same. This not the case, and the indices very much matter. Below are some relevant test cases from the OpenModelica test suite (which includes the expected output):

`eq4.mo`:
```Modelica
// name: eq4.mo
// keywords:
// status: correct
// cflags:   -d=newInst
//

package P
  constant Integer n = 3;
end P;

model A
  Real x;
  parameter Real y;
equation
  x = y * P.n;
  y = x;
end A;

model B
  A a1[3](y = {1, 2, 3});
  A a2[3](each y = 4);
end B;

// Result:
// class B
//   Real a1[1].x;
//   parameter Real a1[1].y = 1.0;
//   Real a1[2].x;
//   parameter Real a1[2].y = 2.0;
//   Real a1[3].x;
//   parameter Real a1[3].y = 3.0;
//   Real a2[1].x;
//   parameter Real a2[1].y = 4.0;
//   Real a2[2].x;
//   parameter Real a2[2].y = 4.0;
//   Real a2[3].x;
//   parameter Real a2[3].y = 4.0;
// equation
//   a1[1].x = a1[1].y * 3.0;
//   a1[1].y = a1[1].x;
//   a1[2].x = a1[2].y * 3.0;
//   a1[2].y = a1[2].x;
//   a1[3].x = a1[3].y * 3.0;
//   a1[3].y = a1[3].x;
//   a2[1].x = a2[1].y * 3.0;
//   a2[1].y = a2[1].x;
//   a2[2].x = a2[2].y * 3.0;
//   a2[2].y = a2[2].x;
//   a2[3].x = a2[3].y * 3.0;
//   a2[3].y = a2[3].x;
// end B;
// endResult
```

`Each1.mo`:
```Modelica
// name:     Each1
// keywords: Each modifier
// status:   correct
//
// Testcase from Modelica specification.
//
model C
  parameter Real a[3];
  parameter Real d;
end C;
model B
  C c[5](each a={1,2,3},d={1,2,3,4,5});
end B;
// Result:
// class B
//   parameter Real c[1].a[1] = 1.0;
//   parameter Real c[1].a[2] = 2.0;
//   parameter Real c[1].a[3] = 3.0;
//   parameter Real c[1].d = 1.0;
[...]

Issue #126, opened 2018-05-25
Nested arrays of components broken


jackvreeken:
In `ArrayExpressions.mo` there currently is a work around for nested indexing not working. Distilled down to its essentials:

```Modelica 
model NestedArrayExpressions
    Real z[3];
end NestedArrayExpressions;
model ArrayExpressions
    NestedArrayExpressions nested2[2];
equation
    nested2[1].z[:] = {4, 5, 6}; // We should also be able to index with `nested2[1].z`
    nested2[2].z[1] = 3;
    nested2[2].z[2] = 2;
    nested2[2].z[3] = 1;
end'ArrayExpressions
```

In general, there's a whole lot more of fancy indexing that we should support, but probably don't. Taken from [xogeny](http://book.xogeny.com/components/subsystems/comp_arrays/#arrays-of-components):

```Modelica 
record Vector3D
  Real x[3];
end Vector3D;

model ArrayExample
  Point p[2];
  Point q[2,3];
  Vector3D v[4];
equation
  p.x = {1, 2}; // p[1].x = 1, p[2].x = 2
  q[:,3].y = {4, 5}; // q[1,3].y = 4, q[2, 3].y = 5;
  q.x = [1, 2, 3; 4, 5, 6]; // q[1,1].x = 1,
                           // q[1,2].x = 2,
                           // q[2,3].x = 6
  v.x[1] = {1, 2, 3, 4}; // v[1].x[1] = 1, v[2].x[1] = 2,
                         // v[3].x[1] = 3, v[4].x[1] = 4
  v[:].x[1] = {1, 2, 3, 4}; // v[1].x[1] = 1, v[2].x[1] = 2,
                            // v[3].x[1] = 3, v[4].x[1] = 4
  v[2:3].x[1] = {2, 3}; // v[2].x[1] = 2, v[3].x[1] = 3
  v[1].x = {1, 2, 3};   // v[1].x[1] = 1, v[1].x[2] = 2,
                        // v[1].x[3] = 3
end ArrayExample;
```

Issue #45, opened 2018-01-23
Cannot pair nested array symbols to their dimensions


jackvreeken:
During flattening we move the dimensions always to the last symbol. This simplifies a lot of things, but does result in not being able to properly name nested arrays symbol later. For a related issue, see #44 .

Example:
```Modelica
model A
    Real x[3];
end A;

model B
    A a[3];
end B;

model C
    B b;
end C;

model D
    C c[2];
end D;

model Test
    D d;
end Test;
```

Results in a symbol `d.c.b.a.x` with dimensions `[2, 3, 3]`. We would like to be able to figure out what dimension belongs to what nested element, such that the user can refer to a symbol named `d.c[1].b.a[2].x[3]` instead of `d.c.b.a.x[1, 2, 3]`. That means we have to store scalar element dimensions in a distinguishable way (see #44), such that we end up with something like `dimensions = [None, 2, None, 3, 3]` or `dimensions = [/1/, 2, /1/, 3, 3]` where `/1/` is of a special (scalar signifying) type.

AnneHommelberg:
fixed by https://github.com/pymoca/pymoca/pull/153/commits/37246e4bffdc79d1c05d1be3a5df83b288b1ee88

Issue #44, opened 2018-01-23
Cannot distinguish between 1-element array and scalar symbols


jackvreeken:
Currently we cannot properly distinguish after parsing/flattening whether a symbol was a 1-element array or a scalar. Example:

```Modelica
model Test
    parameter Integer n = 1;
    Real w[n];
    Real x[1];
    Real y;
    Real z[1,1];
end Test;
```

This matters when we have to name the symbols (e.g. in the CasADi backend when expanding vectors/arrays), such that they can then later be referenced by the user. The user would expect, based on the above Modelica code, that a symbol `x[1]` and a symbol `y` exist. But both have the same signature, which is `sym.dimensions = [1]`.

As it turns out, currently ```symbols['y'].dimensions[0]``` is a `Primary` of with a `value` field of type *int*, whereas `x` is similar but has a *float* inside. So strictly speaking we can distinguish scalars from 1-element arrays, but I would much rather have `.dimensions = []` or `.dimensions = None`, such that we can do a `if not sym.dimensions` in the Python code to detect scalars.

For deeply nested arrays (of arrays, of arrays), I would think of something like `.dimensions = [1, 2, None, 1, None]`. That would however complicate a lot of code, so we could also just opt for the int vs. float approach (or making a custom type "ScalarElementIndex" that is equal to 1). I'm very much open for suggestions.

jackvreeken:
If None does not work well (i.e. results in too many if-else clauses in the `tree.py`), something like below might work well.
```python
class ScalarDimension(int):
    def __new__(cls, *args, **kwargs):
        return  super(ScalarDimension, cls).__new__(cls, 1)
```

AnneHommelberg:
fixed by https://github.com/pymoca/pymoca/pull/153/commits/37246e4bffdc79d1c05d1be3a5df83b288b1ee88

PR #129, opened 2018-05-25
WIP: Fix nested array indexing

jackvreeken:
Indexing an element nested2[1].x, with x itself an array, did not work
properly. No errors were raised, but the result was not equivalent to
using nested2[1].x[:].

To fix this, we now keep track of the dimensions of both nested2 and x
separately, and also how they are indexed. In the above outlined
example, we detect that x is an unindexed array, and we add the slicing
operator [:] ourselves.

Keeping track of dimensions also enables us to detect wrongly indexed
variables, e.g. a user trying to index a scalar, or not specifiying the
right number of dimensions.

To be fixed/addressed in this PR:
- [ ] #126 and all its test cases/examples
- [ ] #127 and all its test cases/examples
- [ ] #44 
- [ ] #45

### DIFF
--- a/src/pymoca/tree.py
+++ b/src/pymoca/tree.py
@@ -487,6 +487,13 @@ def flatten_symbols(class_: ast.InstanceClass, instance_name='') -> ast.Class:
 
             # carry class dimensions over to symbols
             for flat_class_symbol in flat_sub_class.symbols.values():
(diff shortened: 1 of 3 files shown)


jackvreeken:
Mostly addressed by #153

PR #136, opened 2018-06-12
Small fixes

jackvreeken:
- [x] Add/change test case for outputs and expansion

### DIFF
--- a/test/gen_casadi_test.py
+++ b/test/gen_casadi_test.py
@@ -913,7 +913,7 @@ class GenCasadiTest(unittest.TestCase):
         ref_model.alg_states[0].max = 2
         ref_model.alg_states[0].nominal = 1
         ref_model.inputs = list(map(Variable, []))
-        ref_model.outputs = list(map(Variable, []))
+        ref_model.outputs = []
         ref_model.constants = list(map(Variable, []))
         constant_values = []
         for _cst, v in zip(ref_model.constants, constant_values):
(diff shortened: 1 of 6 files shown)


test/gen_casadi_test.py
```diff
@@ -658,7 +658,7 @@ def test_delay_for_loop_with_expand_vectors(self):
         ref_model = Model()
 
         def _array_mx(name, n):
-            return np.array([ca.MX.sym("{}[{}]".format(name, i)) for i in range(n)])
+            return np.array([ca.MX.sym("{}[{}]".format(name, i+1)) for i in range(n)])
```
vdwees: should be `i + 1` tho I am not sure we care at this point

test/models/Annotations.mo
```diff
@@ -0,0 +1,11 @@
+model A
+    Real x, y;
```
vdwees: Why not include an inline annotation here as well?

jackvreeken: I couldn't come up with one that makes sense. I just copied the large complicated annotation from a model that failed.

test/gen_casadi_test.py
jackvreeken: We have `i+1` in `model.py`, so in a sense it's consistent. If I'm going to change this, I'd have to change the formatting of those as well.

vdwees: sure, ok. tox will complain tho I think.

test/models/Annotations.mo
vdwees: fair enough. good for now.

PR #137, opened 2018-06-12
Faster reading cache

jackvreeken:
- [x] Merge #136  first. This goes on top.

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -191,6 +191,45 @@ def save_model(model_folder: str, model_name: str, model: Model,
         for key in ['states', 'der_states', 'alg_states', 'inputs', 'parameters', 'constants']:
             db[key] = [e.to_dict() for e in getattr(model, key)]
 
+
+        # Caching using CasADi functions will lead to constants seemingly
+        # depending on MX variables. Figuring out that they do not is slow,
+        # especially when doing it on a lazy function call, as would be the
+        # case when reading from cache. So instead, we do the depency check
+        # once when saving the model.
+
+        # Metadata dependency checking
+        parameter_vector = ca.veccat(*[v.symbol for v in model.parameters])
+
+        for k, key in enumerate(['states', 'alg_states', 'inputs', 'parameters', 'constants']):
+            metadata_shape = (len(getattr(model, key)), len(CASADI_ATTRIBUTES))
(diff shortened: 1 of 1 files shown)


src/pymoca/backends/casadi/api.py
```diff
@@ -308,19 +346,29 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
...
+                    actual_deps = {all_symbols[j] for j in duration_dependencies[i]}
+                    false_deps = deps - actual_deps
+
+                    if false_deps:
                         [dur] = ca.substitute(
                             [dur],
-                            [false_deps],
-                            [ca.DM(np.full(false_deps.size1(), np.nan))])
+                            list(false_deps),
+                            [ca.repmat(np.nan, *d.size()) for d in false_deps])
```
vdwees: Not 100% sure, but I think that the numpy call is faster, and that you can use it without casting to DM. But maybe there is a good reason to use repmat?

src/pymoca/backends/casadi/api.py
```diff
@@ -191,6 +191,45 @@ def save_model(model_folder: str, model_name: str, model: Model,
...
+        if model.delay_states:
+
+            all_symbols = [model.time,
+                           *model._symbols(model.states),
+                           *model._symbols(model.der_states),
+                           *model._symbols(model.alg_states),
+                           *model._symbols(model.inputs),
+                           *model._symbols(model.constants),
+                           *model._symbols(model.parameters)]
+            all_hashes = [hash(x) for x in all_symbols]
```
vdwees: it is not immediately obvious why you are using a hash. I imagine it is faster to do a linear search on? Why not use an an ordered dict or something else with fast lookups?

src/pymoca/backends/casadi/api.py
jackvreeken: image

repmat is faster than what used to be there. Numpy arrays are 10x faster, but I think CasADi will convert to DM internally. And either way it's likely that `substitute` will be the bottleneck.

vdwees: agreed.

src/pymoca/backends/casadi/api.py
vdwees: Something like
```python
{x: i for i, x in enumerate(all_symbols)}
```

jackvreeken: Fixed.

PR #139, opened 2018-06-20
Various speedups

jackvreeken:
- [x] Merge #138 first (this goes on top)
- [x] Variable.aliases should be a set as well if we do not guarantee order of canonical variables and its aliases
- [x] Update tests with aliases to compare sets instead of list
- [ ] Test with many RTC-Tools models

### DIFF
--- a/src/pymoca/backends/casadi/alias_relation.py
+++ b/src/pymoca/backends/casadi/alias_relation.py
@@ -143,9 +143,14 @@ class AliasRelation:
     def aliases(self, a):
         if self.__is_negative(a):
             a = self.__toggle_sign(a)
-            return OrderedSet([self.__toggle_sign(v) for v in self._aliases.get(a, OrderedSet([a]))])
+            return OrderedSet([self.__toggle_sign(v) for v in self.aliases(a)])
         else:
-            return self._aliases.get(a, OrderedSet([a]))
+            # Instead of the dictionary's "get" method, we avoid always
+            # instatiating an OrderedSet by explicitly checking
+            if a in self._aliases:
+                return self._aliases[a]
+            else:
+                return OrderedSet([a])
 
     def canonical_signed(self, a):
         if self.__is_negative(a):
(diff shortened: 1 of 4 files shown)


jackvreeken:
@jbaayen @vdwees : Comments would be much welcomed

Notable "breaking" changes that I would like comments on:
- `canonical_variables` is now a set instead of an OrderedSet
- `aliases()` returns a set instead of an OrderedSet
- `__iter__`   no longer loops in a guaranteed order (because of the OrderedSet -> set change)

Major difference between sets and OrderedSets is (1) order and (2) indexibility (`my_set[5]`).

Other discussions:
- I find it somewhat confusing that `__iter__` returns the aliases _without_ the canonical state, whereas there is no such function to ask for something like that directly yourself (before, and now). You would have to drop the canonical state from whatever aliases() returns yourself.

jbaayen:
Did you check that dropping guaranteed order does not break things elsewhere? 

Typical use case: Loop over canonical variables, creating an optimization variable vector. Post-optimization, loop again and extract. If we can get two different orderings here, simple use cases like this will break.

jackvreeken:
I haven't encountered any test cases that fail. The AliasRelation is typically not used to loop over. The only case I could find was in `model.py`. On the RTC-Tools side open would loop over the dae's states/controls/etc (which remains the same as before), and possibly over AliasDicts. Note that AliasDicts do not provide any order guarantees when looping either.

I am not against waiting a little with this speed-up until we have more models/real-life use cases.

Keeping order is relatively simple by using OrderedDicts instead (which have a fast C implementation in recent version of CPython), and not assigning the values. It's a bit weird to return an instance of `odict_keys` instead of `set` though.

jbaayen:
I remember that when I first implemented this, I spent a substantial amount time debugging until I found out that there were some order assumptions in RTC. That's when I switched to OrderedDict.

Since we call it an Alias**Relation** I'm not opposed to dropping order, but we need to be sure we're not breaking anything.

PR #144, opened 2018-08-01
Fix delay for loops

jackvreeken:


test/gen_casadi_test.py
```diff
@@ -660,20 +662,22 @@ def test_delay_for_loop_with_expand_vectors(self):
...
+        eps = _array_mx("eps", 1)
 
         ref_model.alg_states = list(map(Variable, [*x, *y, *a]))
+        ref_model.parameters = list(map(Variable, [*eps]))
         ref_model.inputs = list(map(Variable, [*at3_delayed, *z, delay_time]))
         ref_model.inputs[-1].fixed = True
-        # TODO: "a" is not yet deteced as an alias of "x" when expanding.
-        ref_model.equations = [*(x - 5 * z), *(y - at3_delayed), *(a - x)]
+        # TODO: "a" is not yet detected as an alias of "x" when expanding.
+        ref_model.equations = [*(x[1:3] - 5 * z[1:3] * eps), *(y[1:3] - at3_delayed), *(a - x)]
```
AnneHommelberg: Shouldn't this be x[2:3], z[2:3] and y[2:3] since the model now states "for i in 2:3 loop"?

jackvreeken: Modelica is 1-based, and Python is 0-based. Furthermore, Python uses exclusive indexing for the end, whereas Modelica uses inclusive indexing.

PR #153, opened 2018-08-21
Fix indexing 2

AnneHommelberg:
Second attempt to fix the issues concerning array indexing. 

To be fixed/addressed in this pull request:

- [x] #44 
- [x] #45 
- [x] #126 
- [ ] #127

### DIFF
--- a/src/pymoca/backends/casadi/generator.py
+++ b/src/pymoca/backends/casadi/generator.py
@@ -596,43 +596,53 @@ class Generator(TreeListener):
             return float
 
     def get_shape(self, tree):
-        return [self.get_integer(d) for d in tree.dimensions]
+        return [[self.get_integer(d) for d in d_list] for d_list in tree.dimensions]
 
     def get_symbol(self, tree):
         # Create symbol
         shape = self.get_shape(tree)
 
-        if any(isinstance(x, slice) for x in shape):
+        if any(isinstance(x, slice) for var_shape in shape for x in var_shape):
             # Symbol has unspecified dimensions. Value is specified, and
             # carries the correct dimensions.
 
             # We should only get slices as dimensions for a symbol if one of
             # the dimensions is unspecified, i.e. None.
-            assert None in itertools.chain.from_iterable((x.start, x.stop) for x in shape if isinstance(x, slice))
+            assert None in (itertools.chain.from_iterable((x.start, x.stop)
+                            for var_shape in shape for x in var_shape if isinstance(x, slice)))
 
             val_shape = np.array(self.src[tree.value]).shape
 
             # Check if specified dimensions agree between definition and value
-            for dim_i, dim_size in enumerate(shape):
-                if isinstance(dim_size, slice):
-                    continue
+            val_dim_i = -1
+            for var_i, var_shape in enumerate(shape):
+                for dim_i, dim_size in enumerate(var_shape):
+                    if dim_size is None:
+                        continue
 
-                if val_shape[dim_i] != dim_size:
-                    raise Exception("Dimension {} of definition and value for symbol {} differs: {} != {}".format(
-                        dim_i + 1, tree.name, dim_size, val_shape[dim_i]))
+                    val_dim_i += 1
+                    if isinstance(dim_size, slice):
+                        shape[var_i][dim_i] = val_shape[val_dim_i]
+                        continue
 
-            shape = val_shape
+                    if val_shape[val_dim_i] != dim_size:
+                        raise Exception("Dimension {} of definition and value for symbol {} "
+                                        "differs: {} != {}"
+                                        .format(val_dim_i + 1, tree.name, dim_size,
+                                                val_shape[val_dim_i]))
 
-        if len(shape) > 2:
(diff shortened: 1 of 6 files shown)


src/pymoca/backends/casadi/generator.py
```diff
@@ -596,43 +596,51 @@ def get_python_type(tree):
...
 
             # Check if specified dimensions agree between definition and value
-            for dim_i, dim_size in enumerate(shape):
-                if isinstance(dim_size, slice):
-                    continue
-
-                if val_shape[dim_i] != dim_size:
-                    raise Exception("Dimension {} of definition and value for symbol {} differs: {} != {}".format(
+            val_dim_i = -1
+            for var_i, var_shape in enumerate(shape):
```
jackvreeken: I think only one `var_shape` is allowed to have unspecified dimension(s), which could make things a bit simpler. I am not opposed to keeping the logic this way though, but a comment might be helpful.

src/pymoca/backends/casadi/generator.py
```diff
@@ -596,43 +596,51 @@ def get_python_type(tree):
...
+                        continue
+
+                    val_dim_i += 1
+                    if isinstance(dim_size, slice):
+                        shape[var_i][dim_i] = val_shape[val_dim_i]
+                        continue
+
+                    if val_shape[val_dim_i] != dim_size:
+                        raise Exception("Dimension {} of definition and value for symbol {} differs: {} != {}".format(
                         dim_i + 1, tree.name, dim_size, val_shape[dim_i]))
```
jackvreeken: The error message should also print `val_shape[val_dim_i]`, not `val_shape[dim_i]`. Also, `dim_i + 1` might not really make sense anymore either. Would be nice to rethink this  error message such that it provides useful information in the new way of handling things.

src/pymoca/backends/casadi/model.py
```diff
@@ -373,10 +373,27 @@ def simplify(self, options):
                     # For delayed states we do not have any reliable shape
                     # information available due to it being an arbitrary
                     # expression, so we just always expand.
-                    if (old_var.symbol._modelica_shape != (1,) or old_var.symbol.name() in self.delay_states):
+                    if (old_var.symbol._modelica_shape != ((None,),) or old_var.symbol.name() in self.delay_states):
                         expanded_symbols = []
-                        for ind in np.ndindex(old_var.symbol._modelica_shape):
-                            component_symbol = ca.MX.sym('{}[{}]'.format(old_var.symbol.name(), ",".join(str(i+1) for i in ind)))
+
+                        if old_var.symbol.name() in self.delay_states:
```
jackvreeken: A comment explaining what you're doing here (readying the formatted string) would be helpful. I knew because you told me, but otherwise I would have probably been very confused.

test/gen_casadi_test.py
```diff
@@ -27,7 +27,7 @@ class GenCasadiTest(unittest.TestCase):
     def assert_model_equivalent(self, A, B):
         def sstr(a): return set([str(e) for e in a])
 
-        for l in ["states", "der_states", "inputs", "outputs", "constants", "parameters"]:
+        for l in ["alg_states", "states", "der_states", "inputs", "outputs", "constants", "parameters"]:
```
jackvreeken: Could you make adding this the first commit?

test/gen_casadi_test.py
```diff
@@ -1526,13 +1526,164 @@ def test_unspecified_dimensions(self):
                                      [3, 4],
                                      [5, 6]])
 
+    def test_wrong_unspecified_dimensions(self):
```
jackvreeken: This test did not catch the `dim_i` versus `val_dim_i` mismatch in the error message.

jackvreeken:
Parameters/values do not seem to be converted to their proper types anymore with this WIP. I'm getting "MX(0)" instead of a float with value 0.0.

AnneHommelberg:
@jackvreeken Can you provide an example of a model where parameters/values are no longer converted to there proper types? 

That is the last todo for this WIP to fix #44, #45 and #126. Do you still want to merge after that and make a new branch / pull request for #127?

jackvreeken:
The example case is the pumping_station/basic example of RTC-Tools Hydrualic Structures. If you run that with the latest RTC-Tools/RTC-CF/RTC-HS and this branch of Pymoca, it will fail with 

```
Traceback (most recent call last):
  File ".\example.py", line 182, in <module>
    run_optimization_problem(Example, base_folder='..')
  File "path\to\src\rtc-tools\src\rtctools\util.py", line 91, in run_optimization_problem
    prob.optimize()
  File "path\to\src\rtc-tools\src\rtctools\optimization\goal_programming_mixin.py", line 884, in optimize
    self.pre()
  File "path\to\src\rtc-tools-hydraulic-structures\src\rtctools_hydraulic_structures\pumping_station_mixin.py", line
519, in pre
    hr = self._head_range[p.head_option][ps.symbol]
KeyError: MX(0)
```

AnneHommelberg:
@jackvreeken I've found and fixed the problem that caused the MX(0) etc. 
Can you review the changes?
#44, #45 and #126, should all be fixed.

PR #158, opened 2018-09-26
Coerce values to Python type when expanding

jackvreeken:
When not expanding, many symbols are already their Python type if they
can be converted to it. When expanding however, we should take care to
do the same. For example, a ca.DM or ca.MX matrix with a size larger
than one could easily allow us to convert its values to float as well.


AnneHommelberg:
The code looks good but I haven't been able to find a use case for this.
Could a unit test be added with an example of a model for which the expanded symbols were not converted to python type before?
The only unit test (in gen_casadi_test) that passes the python type conversion is already converted to the correct python type in the master branch (test_array_3d).

AnneHommelberg:
@jackvreeken Any update on this?

jackvreeken:
Superseded by #203

PR #161, opened 2018-09-29
Coerce lists stored as arrays to MX

### LINKED ISSUE #160: Arrays in equations cause casadi NotImplementedError
Doing this in modelica causes an error:
```modelica
Deltares.ChannelFlow.Hydraulic.BoundaryConditions.Discharge QBC_OM(M = {QBC_OM_Q * QBC_OM_C})
```
Here we define an equation where M is an array variable of length 1 and we set that to an array expression of length 1: ``M = {QBC_OM_Q * QBC_OM_C}``

The error is as follows:
```
File "c:\users\jvanderwees\dev\pymoca\src\pymoca\backends\casadi\generator.py", line 349, in exitEquation
    src_right = ca.MX(src_right)
File "C:\Users\jvanderwees\dev\polders\bos-rijnland-polders\venv\lib\site-packages\casadi\casadi.py", line 10542, in __init__
    this = _casadi.new_MX(*args)

NotImplementedError: Wrong number or type of arguments for overloaded function 'new_MX'.
  Possible prototypes are:
    MX()
    MX(Sparsity)
    MX(float)
    MX([float])
    MX(DM)
    MX(MX)
    MX(int,int)
    MX(Sparsity,MX)
  You have: '([MX])'
```

This is because we store arrays (the right hand side, in this case) as lists in `self.src`, and don't have any handling to coerce them to `ca.MX` objects. The solution would be to
1) store them as `ca.MX` in `self.src` so that `get_mx()` actually returns a `ca.MX` object
2) coerce the list to `ca.MX` on an ad-hoc basis

vdwees:
Arrays are stored as lists in self.src, but need to be converted to
ca.MX types when we construct equations. Here we detect and do the
coersion using ca.vertcat()

Closes #160

### DIFF
--- a/src/pymoca/backends/casadi/generator.py
+++ b/src/pymoca/backends/casadi/generator.py
@@ -336,11 +336,15 @@ class Generator(TreeListener):
             src_left = ca.vertcat(*[self.get_mx(c) for c in tree.left])
         else:
             src_left = self.get_mx(tree.left)
+            if isinstance(src_left, list):
+                src_left = ca.vertcat(*src_left)
 
         if isinstance(tree.right, list):
             src_right = ca.vertcat(*[self.get_mx(c) for c in tree.right])
         else:
             src_right = self.get_mx(tree.right)
+            if isinstance(src_right, list):
+                src_right = ca.vertcat(*src_right)
 
         src_left = ca.MX(src_left)
         src_right = ca.MX(src_right)


jackvreeken:
Could you add a test where this failed (and now succeeds)? Generally, CasADi has happy with np.ndarray in expressions, just not lists. So converting to np.ndarray might be safer if the argument can be a nested list (e.g. `[[1,2,3],[4,5,6]]`).

PR #164, opened 2018-11-01
Implement logical operators not, and, or, in CasADi generator.

jbaayen:
- [x] Add support in parser


jbaayen:
I am unsure btw whether to use "1 - x" as "not" expression, or "if_else(x, 0, 1)". Currently I've put in "1 - x" as it results in simpler expressions.

jackvreeken:
Something like ```e = not (1 or 1 or 0);``` will fail with "not" implemented like that though. I'm not sure if it's prettier to make sure that `or` always results in 1 or 0, or that `not` works in a way that allows "true" values to be anything other than zero.

jbaayen:
True. Let's use `if_else` then, as it's more robust.

PR #166, opened 2018-11-22
casadi.Model:  Recursify simplify/eliminable_variable_expressions

jbaayen:
Also added support for if/else expressions.  These changes allow many additional states to be eliminated in RTC-Tools Channel Flow.


javdejong:
Could use a test that verifies (and shows) new behavior.

src/pymoca/backends/casadi/model.py
```diff
@@ -500,17 +500,17 @@ def simplify(self, options):
...
-                    continue
+                    return eq, 0.0
+
+                if eq.is_op(ca.OP_IF_ELSE_ZERO):
+                    cond = eq.dep(0)
+                    nested_eq = eq.dep(1)
+
+                    variable, value = extract_assignment(nested_eq)
+                    if variable is not None:
+                        return variable, ca.if_else(cond, value, 0, True)
```
jackvreeken: Why the short-circuiting?

src/pymoca/backends/casadi/model.py
```diff
@@ -524,16 +524,42 @@ def simplify(self, options):
...
                         if eq.is_op(ca.OP_SUB):
-                            values.append(value)
+                            return variable, value
                         else:
-                            values.append(-value)
+                            return variable, -value
 
-                        # Skip this equation
-                        continue
+                if eq.n_dep() == 2 \
```
jackvreeken: `elif`? Also, wrapping with parentheses is generally preferred to using `\`

src/pymoca/backends/casadi/model.py
```diff
@@ -524,16 +524,42 @@ def simplify(self, options):
...
+                    and eq.is_op(ca.OP_ADD) \
```
jackvreeken: Can this also be generalized to OP_SUB?

src/pymoca/backends/casadi/model.py
```diff
@@ -524,16 +524,42 @@ def simplify(self, options):
...
+                    cond_1 = eq.dep(0).dep(0)
+                    cond_2 = eq.dep(1).dep(0)
+                    nested_eq_1 = eq.dep(0).dep(1)
+                    nested_eq_2 = eq.dep(1).dep(1)
+
+                    variable_1, value_1 = extract_assignment(nested_eq_1)
+                    variable_2, value_2 = extract_assignment(nested_eq_2)
+
+                    if variable_1 is not None and variable_2 is not None and variable_1.name() == variable_2.name():
+                        return variable_1, ca.if_else(cond_1, value_1, 0, True) + ca.if_else(cond_2, value_2, 0, True)
```
jackvreeken: 1) If this works for e.g. an `x + y`, will it also work for `z + x +y` (with all of them assigning to the same variable)
2) What if z is a different variable or cannot be resolved. There is no op/dep that matches just x + y, there is only one that does 'z + x` and one that does `(z + x) + y`

src/pymoca/backends/casadi/model.py
jbaayen: We short-circuit everywhere in generator.py. If we would not short-circuit here, we would change the short-circuiting behaviour of the processed if/else statements.

src/pymoca/backends/casadi/model.py
jbaayen: The reason it is set up like this is that casadi if_else() produces two if_else_zero() ops and adds them up. I will add a comment.

src/pymoca/backends/casadi/model.py
jbaayen: See my response to the comment above.

src/pymoca/backends/casadi/model.py
jackvreeken: How do we end up with a `ca.IF_ELSE_ZERO` then? 

```Python
x = ca.MX.sym("x")
y = ca.MX.sym("y")
cond = ca.MX.sym("cond")

eq = ca.if_else(cond, y - x, 0, True)
eq.is_op(ca.OP_IF_ELSE_ZERO)  # --> False

eq2 = ca.if_else(cond, y - x, 0)
eq2.is_op(ca.OP_IF_ELSE_ZERO)  # --> True
```

jbaayen:
```
eq = ca.if_else(cond, y, x, True)                                                                                                                     
[eq] = _expand_simplify_mx([eq])                                                                                                                      

eq.is_op(ca.OP_ADD)  # True                                                                                                                            
eq.dep(0).is_op(ca.OP_IF_ELSE_ZERO)  # True                                                                                                              
eq.dep(1).is_op(ca.OP_IF_ELSE_ZERO)  # True                                                                                                     
```

jackvreeken: Ah, so that's how.

src/pymoca/backends/casadi/model.py
```diff
@@ -524,16 +524,44 @@ def simplify(self, options):
...
+                            return variable, -value
 
-                        # Skip this equation
-                        continue
+                if eq.n_dep() == 2 \
+                    and eq.is_op(ca.OP_ADD) \
+                    and eq.dep(0).is_op(ca.OP_IF_ELSE_ZERO) \
+                    and eq.dep(1).is_op(ca.OP_IF_ELSE_ZERO):
+                    # Once passed through _expand_simplify_mx, a CasADi if_else
+                    # is converted to a sum of two if_else_zero operations.
```
jackvreeken: Good comment. This should make it clear enough that this only applies in the expand_mx case (which is the only time expand_simplify_mx is called afaik).

test/gen_casadi_test.py
```diff
@@ -1158,6 +1158,35 @@ def test_simplify_eliminable_variable_expression(self):
         # Compare
         self.assert_model_equivalent_numeric(casadi_model, ref_model)
 
+    def test_simplify_eliminable_variable_expression_if_else(self):
```
jackvreeken: Good and clear example. Thanks!

PR #167, opened 2018-11-22
Eliminable differentiated states

jbaayen:
This is, again, useful for RTC-Tools Channel Flow, so that we can eliminate differentiated states from optimization problems if desired.

### DIFF
--- a/src/pymoca/backends/casadi/model.py
+++ b/src/pymoca/backends/casadi/model.py
@@ -513,10 +513,17 @@ class Model:
 
             p = re.compile(options['eliminable_variable_expression'])
 
+            states = OrderedDict([(s.symbol.name(), s) for s in self.states])
             alg_states = OrderedDict([(s.symbol.name(), s) for s in self.alg_states])
 
+            all_states = OrderedDict([(s.symbol.name(), s) for s in itertools.chain(self.states, self.alg_states)])
+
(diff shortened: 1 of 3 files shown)


jackvreeken:
You probably already figured this out, but the tests are failing on the f-format string (not supported yet in Python 3.5). You would have to use one of the older ways of formatting strings.

src/pymoca/backends/casadi/model.py
```diff
@@ -568,21 +580,61 @@ def extract_assignment(eq):
...
+
             reduced_equations = []
             for eq in self.equations:
                 variable, value = extract_assignment(eq)
                 if variable is not None:
+                    if variable.name() in states:
+                        # Mark derivative state for replacement too.
+                        derivative = get_derivative(value)
+
+                        state_index = list(states.keys()).index(variable.name())
```
jackvreeken: Using `.index` could be dangerously slow, and will likely come back to bite us. It's probably better to make a dict `MX state -> MX derivative` or `str state (name) -> MX derivative`. Note that all other lookups like `eq.dep(0).name() in states` are already O(1).

jackvreeken:
@jbaayen Pushed one commit on top to get rid of the `.index`. Could you check if you agree with my implementation?

jbaayen:
Useful improvement -- thanks! Looks good.

PR #168, opened 2018-11-23
Fix infinite loop in replace_parameter_expressions

jackvreeken:
In the replace_parameter_expressions simplification step, it was possible
to end up in an endless loop when there are a lot of parameters that have
symbolic values. With the expansions of vectors first (when expand_mx is
True) introduced with 028093f, it is more likely to end up with such
cases.

The CasADi ca.is_equal call will terminate ehen its maximum depth is
reached, and return False. The limit of 100 introduced in adbd468e is not
enough in practice to compare two MX graphs. We therefore increase this
number significantly.

A safeguard is also introduced to prevent an infinite loop from happening
again, by raising an Exception after (currently) 100 iterations.

### DIFF
--- a/src/pymoca/backends/casadi/model.py
+++ b/src/pymoca/backends/casadi/model.py
@@ -4,6 +4,7 @@ import numpy as np
 import itertools
 import logging
 import re
+import sys
 
 from pymoca import ast
 from .alias_relation import AliasRelation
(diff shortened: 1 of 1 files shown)


src/pymoca/backends/casadi/model.py
```diff
@@ -11,7 +11,8 @@
 
 logger = logging.getLogger("pymoca")
 
-CASADI_COMPARISON_DEPTH = 100
+CASADI_COMPARISON_DEPTH = 1000000
```
jbaayen: use `sys.maxsize`?

jackvreeken: Well, I think even 1 million is a bit large. Maybe 1000 or 10k is high enough. What I want to avoid is infinite loops, and sys.maxsize will be a very long wait leading to users blaming pymoca for being slow. Instead, we should just raise an Exception indicating that something is wrong with the model.

Note that the reason for these deep graphs (with RTC-CF models) is CasADi's implementation of `linspace`. I have a pending PR on their side addressing this: https://github.com/casadi/casadi/pull/2338

jbaayen: I always shudder when I see arbitrary constants ...

PR #169, opened 2018-11-23
Speed ups for uncaching models with delayed states

jackvreeken:
- [x] Check that results are _exactly_ equal
- [x] Merge !168 first

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -366,19 +366,39 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
 
             duration_dependencies = db['__delay_duration_dependent']
 
+            # Get rid of false dependency symbols not used in any delay
+            # durations. This significantly reduces the work the (slow)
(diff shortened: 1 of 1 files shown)


src/pymoca/backends/casadi/api.py
```diff
@@ -366,19 +366,39 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
 
             duration_dependencies = db['__delay_duration_dependent']
 
+            # Get rid of false dependency symbols not used in any delay
+            # durations. This significantly reduces the work the (slow)
+            # substitute() calls have to do later on.
```
jbaayen: Some additional explanation as to why this is the case would be helpful.

jackvreeken: Why we get those false dependencies you mean, or why we want to get rid of them?

jbaayen: Reading this code, I was left with the following question: Why does getting rid of the false dependencies have a significant impact on performance?

PR #171, opened 2018-12-08
Fix initialization of derivative symbol in some cases

### LINKED ISSUE #163: broken assertion
In this block of code in generator.py:
``` python
def get_indexed_symbol(self, tree, s):
       assert len([dim for shape in s._modelica_shape for dim in shape if dim is not None]) <= 2,\
       "Dimensions higher than two are not yet supported"
```
I get the error `TypeError: 'int' object is not iterable`.  Perhaps the way we store shapes has changed, and the assertion is out of date? Note that in my case, `s._modelica_shape` is `(3, 1)`.

Line in question: https://github.com/pymoca/pymoca/blob/master/src/pymoca/backends/casadi/generator.py#L705

jackvreeken:
If the first encountered expression of a certain derivative was something
like "der(x[1])", the symbol for the derivative was not properly
inititialized.

Note that if the first encountered expression was instead something like
"der(x)" or "der(x[i])" (like in for loops), the initialization was done
correctly. Encountering an expression like "der(x[1])" afterwards would
then also not lead to any problems.

Closes #163

### DIFF
--- a/src/pymoca/backends/casadi/generator.py
+++ b/src/pymoca/backends/casadi/generator.py
@@ -689,7 +689,10 @@ class Generator(TreeListener):
             dep = s.dep()
             if dep.name() not in self.derivative:
                 der_dep = _new_mx("der({})".format(dep.name()), dep.size())
+                der_dep._modelica_shape = \
+                    self.nodes[self.current_class][dep.name()]._modelica_shape
                 self.derivative[dep.name()] = der_dep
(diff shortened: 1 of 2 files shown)


jackvreeken:
Also needs to be merged to the maintenance/0.3.x branch.

jackvreeken:
Ah yes, you are right. We should open a separate issue about that.

PR #176, opened 2018-12-12
Fix naming of derivatives when expanding vectors

### LINKED ISSUE #173: Expand vectors results in wrong names for derivatives
`der(a.x)` with both `a` and `x` a 1D-array turns into names like `der(a[1].x)[2]`.

jackvreeken:
Closes #173


jackvreeken:
@AnneHommelberg A quick sanity check would be appreciated. It passes all tests (and the new test I added), so I don't expect anything to be wrong. I chose the regular expression over "split" and "find", because I ended up with a lot of if-else blocks. 
Note that the regular expression supports the possibillity of "der(der(", but that does not work currently (and I'm not sure if it's even allowed by the spec).

PR #179, opened 2019-01-09
Proposal to add two custom builtin functions to do 1D and 2D interpolation

jbaayen:
Interpolation is not officially supported by the Modelica core, whence the proposed builtins are prefix with the string "_pymoca_": _pymoca_interp1d and _pymoca_interp2d.

### DIFF
--- a/src/pymoca/backends/casadi/generator.py
+++ b/src/pymoca/backends/casadi/generator.py
@@ -6,7 +6,7 @@ from collections import namedtuple, deque, OrderedDict
 import casadi as ca
 import numpy as np
 import itertools
-from typing import Union, Dict
+from typing import Union, Dict, Iterable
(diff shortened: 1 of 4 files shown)


jackvreeken:
I haven't looked into it too much, but `Modelica.Math.Vectors.interpolate` is not the type of interpolation you're looking for? Otherwise I'm okay with your changes.

jbaayen:
The implementation of `Modelica.Math.Vectors.interpolate` in the standard library heavily relies on `while`, which does not map well into CasADi space. The standard library also does not support B-Spline interpolation, which you would want to be able to do continuous optimization.

Shall we just go for these custom functions then? We can always provide pure Modelica implementations for compatibility with other tools if needed.

PR #182, opened 2019-02-22
Tree: Fix handling of constant input value attribute

jackvreeken:
Just like parameters and constants, the value attribute on a constant
input should not be replaced by an equation; it should be left as-is.

- [ ] Breaks/incompatible with the 'InlineAssignment' model
- [ ] Add test to catch this bug
- [ ] Also merge to stable branch

Thinking about it, I would think the 'InlineAssignment' model would be better formulated like 
```Modelica
model InlineAssignment
    Real x = y;
    input Real y(fixed=true);
end InlineAssignment;
```
Setting a constant input equal to a state does not make sense to me, but the reverse does. Not quite sure if we can catch the difference though.

### DIFF
--- a/src/pymoca/tree.py
+++ b/src/pymoca/tree.py
@@ -991,7 +991,8 @@ def add_state_value_equations(node: ast.Node) -> None:
     non_state_prefixes = {'constant', 'parameter'}
     for sym in node.symbols.values():
         if not (isinstance(sym.value, ast.Primary) and sym.value.value is None):
-            if len(non_state_prefixes & set(sym.prefixes)) == 0:
+            is_const_input = 'input' in sym.prefixes and isinstance(sym.fixed, ast.Primary) and sym.fixed.value is True
+            if len(non_state_prefixes & set(sym.prefixes)) == 0 and not is_const_input:
                 node.equations.append(ast.Equation(left=sym, right=sym.value))
                 sym.value = ast.Primary(value=None)
 


jackvreeken:
@jbaayen > Does this make sense to you, or should we stop reading "constant inputs" from the Modelica model on the RTC-Tools side?

jbaayen:
> Setting a constant input equal to a state does not make sense to me, but the reverse does.

Modelica is acausal, so it would be wrong if this were to make a difference.

jbaayen:
Have you considered adding "input" to `non_state_prefixes`?

jackvreeken:
Constant inputs are very much an RTC-Tools thing though.

If we would add "input" to `non_state_prefixes`, something like below would end up not working anymore with RTC-Tools (whereas it works now):

```Modelica
model InlineAssignment
    Real x;
    input Real y = x;
end InlineAssignment;
```

It would be acausal as well then, because this would work:
```Modelica
model InlineAssignment
    Real x = y;
    input Real y;
end InlineAssignment;
```

jbaayen:
Good point. This makes me think that the present behaviour is not a bug and should not be changed. 

It would mean that in RTC-Tools, we can no longer override inputs for which a value has been specified. This is arguably correct behaviour as the assignments form a part of the model equations, which cannot be overridden elsewhere either.

jackvreeken:
Fine by me. Just note that there is currently no way to know if a value was specified in the model (which was moved to the equations), so if someone sets another value in RTC-Tools for such a constant input it will fail somewhere (probably the solver saying it's infeasible). No user friendly exception or anything.

PR #188, opened 2019-07-30
Split installation dependencies per backend

jackvreeken:
- [x] Make sure to tag 0.5.x, as this is backwards incompatible. Others will need to expliticly specify the backends they use when relying on us now, e.g . "pymoca[casadi]"  instead of just "pymoca".

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
-include requirements.txt
 include LICENSE
 include README.md
 recursive-include doc *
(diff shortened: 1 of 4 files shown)


jackvreeken:
@gitlyap Is this something you can get behind? I only need the CasADi backend, and do not like always dragging along the sympy and lxml dependencies that come with pymoca. For you with the CasADi backend, you might not need lxml and CasADi (and matplotlib/jupyterlab).

Instead of "pip install pymoca", you would then do "pip install pymoca[sympy]" (or "pip install pymoca[all]" if you want behaviour similar to what happens now). You can also list it as such as a dependency in a setup.py of another project, i.e. "pymoca[casadi] >= 0.4.4".

PR #212, opened 2020-02-10
CasADi: Warn if start attribute of alias conflicts

### LINKED ISSUE #208: detect_aliases overwrite non default starting values without prompting the user
When detecting and eliminating aliases in the casadi model_transfer api the start value of the possibly non-default canonical state start value is ignored as soon as one of the aliases has a non-default start value

``` python
# If any of the aliases has a nondefault start value, apply it to
# the canonical state as well
alias_state_start = ca.MX(alias_state.start)
if not alias_state_start.is_constant() or not alias_state_start.is_zero():
     start = sign * alias_state.start
```

Imho the start value of the canonical variable should have the precedence over the aliases and maybe we should warn the user that non-default start values of the aliases have been ignored. 

We could do something like:

``` python
# If any of the aliases has a nondefault start value, apply it to
# the canonical state as well
alias_state_start = ca.MX(alias_state.start)
# check if the alias has a non-deault start value
if not alias_state_start.is_constant() or not alias_state_start.is_zero():
    # check if a non-default value was already present
    if not ca.MX(start).is_constant() or not ca.MX(start).is_zero():
        # check if the value is not the same
        if start != alias_state.start:
            # prompt the user that alias start value has been ignored
            logger.info("Non-default starting value for alias {} ignored.".format(alias))
    else:
        start = sign * alias_state.start
```

I found this particularly important when using units from the modelica standard library that have a starting values set. 

PS: please forgive me if this is not the way to public issues, I am quite new to cooperating on github. Maybe I should have opened a pull request? I am ready to accept any suggestion!

jackvreeken:
Instead of just blindly overwriting the start attribute with that of the
lastest alias, we should check if there is a conflict to avoid suprises.

Note that we cannot exhaustively check for conflicts, because if the
user sets the canonical state's start attribute explicitly to zero, it
will still overwrite it with that of the alias.

Also note that the order in which we loop over aliases is random (due to
the usage of a set), so the order in which messages appear and what
conflict they report can be different. This also means that, in the case
the user does not set a non-zero start attribute on the canonical state
(that is the first encountered alias), but does do so on multiple of its
aliases, the eventual start attribute of the canonical state cannot be
known in advance. This situation should be prevented by making sure that
all aliases have matching (or not-set) start attributes. The warnings we
added in this commit will help with that.

Closes #208

### DIFF
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -316,8 +316,7 @@ class Symbol(Node):
         self.outer = False  # type: bool
         self.dimensions = [[Primary(value=None)]]  # type: List[List[Union[Expression, Primary, ComponentRef]]]
         self.comment = ''  # type: str
-        # params start value is 0 by default from Modelica spec
-        self.start = Primary(value=0)  # type: Union[Expression, Primary, ComponentRef, Array]
+        self.start = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
         self.min = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
         self.max = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
         self.nominal = Primary(value=None)  # type: Union[Expression, Primary, ComponentRef, Array]
--- a/src/pymoca/backends/casadi/model.py
+++ b/src/pymoca/backends/casadi/model.py
@@ -17,6 +17,10 @@ SUBSTITUTE_LOOP_LIMIT = 100
 CASADI_ATTRIBUTES = [attr for attr in ast.Symbol.ATTRIBUTES if not attr == 'unit']
 
 
+class _DefaultValue(int):
+    pass
+
+
 class Variable:
     def __init__(self, symbol, python_type=float, aliases=None):
         if aliases is None:
@@ -27,7 +31,7 @@ class Variable:
 
         # Default attribute values
         self.value = np.nan
-        self.start = 0
+        self.start = _DefaultValue(0)
         self.min = -np.inf
         self.max = np.inf
         self.nominal = 0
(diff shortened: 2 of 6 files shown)


jackvreeken:
@PeterRabbit95 I think this would fix you issue. Could you check?

Please also check if I managed to explain nuances for this warning in the commit message. It's kind of annoying dealing with the default start attribute value of `0`, and not being able to discern this from an explicitly set 0. There might be a ways around that eventually, but for now I don't really want to think about that as that probably causes headaches in other ways :P

PeterRabbit95:
Hi @jackvreeken I just tested your solution and it works properly with my model! Thanks a lot for your support. I do agree that defining every 0 start value as default is a bit dangerous but this is a step in the right direction :)) 

PS: allow me a little off-topic: I am currently using pymoca quite extensively and I am implementing some optimal control related features. Is there a place (forum ecc.. ) where I could propose/discuss what I would like to achieve so that I can maybe get some guidance or be informed of features that have already been implemented?

jackvreeken:
Good to hear, and thanks for checking! I'll think about doing something for that initial value. I'm thinking something along the lines of

```Python
class DefaultFloatValue(float):
    pass

# ...

class Variable:
    def __init__(self, symbol, python_type=float, aliases=None):
        # ...
        self.start = DefaultFloatValue(0.0)
```

And then replace e.g. 
```Python
alias_state_start.is_constant() or not alias_state_start.is_zero()
```
with 

```Python
isinstance(alias_state_start, DefaultFloatValue)
```
But I'll probably only get around to testing this in ~2 weeks time unfortunately...

As far as a discussion on optimal control is concerned, I'm not really sure what would be a good place to discuss. Ideas are more than welcome, and I would love to hear more about what you are doing! I mostly use Pymoca with [RTC-Tools](https://gitlab.com/deltares/rtc-tools) by the way, for the (model predictive) control of surface and/or pipeline water systems.

PR #215, opened 2020-02-26
Add method to API in order to retrieve the default compiler options

PeterRabbit95:
I am not sure if writing some documentation is a plan for the future. Nevertheless, I think that getting a list of the available options could come quite handy, especially for newcomers. I generated this list based on the get methods used to access the options dictionary.

I could also write a description of each option in the docString if this sounds useful.

### DIFF
--- a/src/pymoca/backends/casadi/_options.py
+++ b/src/pymoca/backends/casadi/_options.py
@@ -1,0 +1,40 @@
+def _get_default_options():
+    """
+    Returns a dictionary with the available compiler options and the default values
+
+    Returns:
+        default_options (Dict)
+
+    """
+    return {
+        'library_folders': [],
+        'verbose': False,
+        'check_balanced': True,
+        'mtime_check': True,
+        'cache': False,
+        'codegen': False,
+        'expand_mx': False,
+        'unroll_loops': True,
+        'inline_functions': True,
+        'expand_vectors': False,
+        'replace_parameter_expressions': False,
+        'replace_constant_expressions': False,
+        'eliminate_constant_assignments': False,
+        'replace_parameter_values': False,
+        'replace_constant_values': False,
+        'eliminable_variable_expression': None,
+        'factor_and_simplify_equations': False,
+        'detect_aliases': False,
+        'reduce_affine_expression': False,
+    }
+
+
+def _merge_default_options(options):
+    if options is None:
+        return _get_default_options()
+    elif isinstance(options, dict):
+        default_options = _get_default_options()
+        default_options.update(options)
+        return default_options
+    else:
+        raise TypeError('options must be of type dict')
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -15,6 +15,7 @@ from pymoca import __version__
 from . import generator
 from .alias_relation import AliasRelation
 from .model import CASADI_ATTRIBUTES, Model, Variable, DelayArgument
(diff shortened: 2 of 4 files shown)


jackvreeken:
I definitely agree this is a good idea. It's very hard to tell right now what simplification steps are enabled by default.

I would suggest also making sure that in model.py you refer to this API call (or merge it as such) to .get the correct default value. Now we do 

```Python
options.get('expand_vectors', False) 
```

Then the 'False' part would fall out. You can refer to the value of the default dict, or in the API do something like

```Python
options = get_default_compiler_options()
options.update(compiler_options)
```
and then replace the above thing in `model.py` with 
```Python
options['expand_vectors']
```
You can also define the default values elsewhere (e.g. in model.py), and then refer to them in the `.get()` call. Either way I would like to avoid having the default value specified twice, because that will go wrong eventually.

PeterRabbit95:
Thanks a lot for your feedback! I have implemented what you suggested. 
To avoid import loops I created a new file dedicated to the options default values.

jackvreeken:
I find it rather ugly how both `model.py`, `generator.py` _and_ `api.py` rely on the compiler_options dictionary. I hadn't seen it this clear before, but to me it feels like the the complete compiler_options dictionary with all its defaults should only really be part of `api.py`, not of the other two. Not sure how to accomplish this though, but first I want to check if you agree.

For the `simplify` method in `model.py` I think we can argue that by default it does nothing. So that means we can go back to using `.get`, but just not explicitly specify the default value. Then we do not need to now anything about any options.

For the generator, it's weird because it does want a bunch of options. Although the type it now assumes is wrong (it's not just `bool` anymore; there are also `str` in there).

Not saying I will hold off on merging this, because Pymoca deliberately still has a somewhat unstable API. But if we do want to work towards it, then I suggest we make `options.py` more private (rename to `_options.py`?), expose the methods in it publicly as well via `api.py`.

I'm also not sure about the necessity of the `merge_default_options` method. I think I'd rather not add that to the public/stable API for now, and just keep it internal. A shorthand (aside from the type check) is `{**get_default_options(), **(compiler_options or {})}`

PeterRabbit95:
Starting from the two final points:
1. I totally agree with you on not exposing options. I should have done that from the beginning but, as you may have noticed, I am still learning the best practices with python. I'll commit that asap. I am sorry but I am not sure I got what you mean here "A shorthand (aside from the type check) is {**get_default_options(), **(compiler_options or {})} by the way, which you may or may not prefer over a function call."
2. I am not sure about this point. The options are shared between the compiler and the simplification process so naming it `get_default_compiler_options` could be misleading. Diving them into 2 different dictionaries anyway seems too much. I think exposing them through the api leaving a generic name might be a good solution.

jackvreeken:
Provided `a` and `b` are dicts:
```Python
c = {**a, **b}
# Is the same as
c = a.copy()
c.update(b)
```
Anyway, it looks good to me now. I have no idea how to push commits to your branch, but I wanted to squash everything into a single commit: https://github.com/pymoca/pymoca/commit/fca3bd7ec0704bcefdcafc6e3fbe02c158011140

I can merge that branch (and close this MR without merging), or you can do the same on your end and I'll "rebase and merge" this MR. Whichever you prefer.

PeterRabbit95:
Hi, thanks for the explanation! I squashed and pushed on my fork, proceed as it's more convenient for you :)

PeterRabbit95:
Sorry, I have just noticed I did not respect the naming standard of the different backends in the commit message... "CasADi:..." I'll be more careful in the future

Issue #214, opened 2020-02-26
multiple calls to model simplify with casadi backend are risky


PeterRabbit95:
Calling model simplify more than once (including the implicit call during model building) can be risky due to issues mainly caused by alias detection. 

Multiple calls to the model simplification process can be beneficial to reduce the number of equations especially when alternating detect_aliases and eliminate_constant_assignments. Unfortunately, this might often raise errors:

One scenario is when two calls of detect_aliases are made.

Consider the following simple model:

```modelica
model AlgStates

Real x1(start = 0);
Real x2(start = 0);
Real z;
Real f;
input Real u(min = -5, max = 10);
input Real u1(min = -5, max = 100);

equation

der(x2) = x1 + u1;
der(x1) = u + z;
f = 0.2;
z = f;

end AlgStates;
```

On the second call to detect_aliases it will raise an error because the variable z won't be amongst the model variables anymore.

A possible solution to this is changing (in model.py)
```python
alias_state = all_states[alias]
```
with
```python
try:
    alias_state = all_states[alias]
    logger.debug("Detecting aliases: merging info from {} to {}".format(alias, canonical))
except KeyError:
    logger.debug("Detecting aliases: alias {} is not available. Probably you are calling simplify 
again".format(alias))
    continue
```

Another scenario is when there is the following call chain: detect_aliases, eliminate_constant_assignments and detect_aliases.

The risk here is that eliminate_constant_assignments eliminates a canonical variable created by the first detect_aliases.

Imho we should decide whether to allow multiple calls to simplify (and in this case, we should improve the robustness, e.g with checks similar to the one I proposed) or whether not to allow them (and in this case, a single call should be more effective)

jackvreeken:
Do you have any preference for which one to prefer? I think until now I just assumed that additional simplify() calls would just be noops, but apparently they are not.

PeterRabbit95:
I did not have time to implement anything but I feel like the easiest path is to make simplify more robust and allow multiple calls. I'll try to play around tomorrow. Furthermore, I would even implement a 'simplify_recursive' option where the user can define a number of calls to be implicitly made. We could then probably check if there is actually any change and in case there isn't stop the process earlier.

PR #216, opened 2020-03-04
Allow multiple calls to simplify and support iterative_simplification option

PeterRabbit95:
As discussed in issue #214, calling simply multiple times might cause errors. This pull request solves 2 errors with detect aliases:

- multiple calls to detect aliases that used to fail because when substituting the variable attributes the second time a keyError is raised
- a call chain like detect_aliases - eliminate_constant_assignments - detect_aliases where eliminate_constant_assignments eliminated a canonical variable

Furthermore an option iterative_simplification is implemented to maximize the simplification of particularly tangled models (for example with many connectors and thus many aliases and trivial equations)

### DIFF
--- a/src/pymoca/backends/casadi/_options.py
+++ b/src/pymoca/backends/casadi/_options.py
@@ -26,6 +26,7 @@ def _get_default_options():
         'factor_and_simplify_equations': False,
         'detect_aliases': False,
         'reduce_affine_expression': False,
+        'iterative_simplification': False
     }
 
 
--- a/src/pymoca/backends/casadi/alias_relation.py
+++ b/src/pymoca/backends/casadi/alias_relation.py
@@ -33,6 +33,9 @@ class AliasRelation:
             self._canonical_variables_map[v] = (canonical_a, sign_a)
             self._canonical_variables_map[self.__toggle_sign(v)] = (canonical_a, -sign_a)
 
+    def remove_canonical_variable(self, a):
+        self._canonical_variables.remove(a)
+
     def __toggle_sign(self, v):
         if self.__is_negative(v):
             return v[1:]
--- a/src/pymoca/backends/casadi/alias_relation.py
+++ b/src/pymoca/backends/casadi/alias_relation.py
@@ -8,6 +8,7 @@ class AliasRelation:
         self._aliases = {}
         self._canonical_variables_map = OrderedDict()
         self._canonical_variables = set()
+        self._substituted_aliases = set()
 
     def add(self, a, b):
         # Construct aliases (a set of equivalent variables)
@@ -61,6 +62,9 @@ class AliasRelation:
             else:
                 return a, 1
 
+    def confirm_substitution(self, a):
+        self._substituted_aliases.add(a)
+
     @property
     def canonical_variables(self):
         return self._canonical_variables
@@ -71,3 +75,7 @@ class AliasRelation:
             aliases = self.aliases(canonical_variable).copy()
             aliases.discard(canonical_variable)
             yield canonical_variable, aliases
+
+    @property
+    def substituted_aliases(self):
+        return self._substituted_aliases
--- a/src/pymoca/backends/casadi/model.py
+++ b/src/pymoca/backends/casadi/model.py
@@ -865,50 +865,53 @@ class Model:
                     else:
                         sign = 1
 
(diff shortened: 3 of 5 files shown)


jackvreeken:
Thanks! I'll get around to properly testing/checking this on Saturday or Sunday.

PeterRabbit95:
I just rebased this on the last version of the master with the separated api for the options. As soon as I did it I realized it has been a bad idea. I am sorry, I should have merged the new version of the master into this branch as you were maybe working on that.

jackvreeken:
I'm starting to understand what's going on and what you fixed, but I need some more time to completely wrap my head around it. And don't worry about the accidental rebase, I'm still in the process of figuring out what is going wrong and why (before I move on to figuring out why your fix works, and whether we may or may not be able to do it better/cleaner/etc).

PeterRabbit95:
Sounds good! I really appreciate your support. Sorry if my solutions are a bit crude, I still struggle to see the full picture of what I do and how they'll affect the rest of the code.

PeterRabbit95:
If you think it could help, I can share with you the model on which I get a big benefit using the interative_simplification. It's something I'd rather not post here so you can drop me a quick mail to pietropelizzari@live.it and I can forward it to you.

jackvreeken:
OK, so I took a look to see if I could figure out what was going wrong. I think it was a rather simple fix in the end (but hard to figure out where to apply said easy fix 😂). And I heavily relied on the work you did of course! Could you take a look at https://github.com/pymoca/pymoca/commits/issue-214 ?

I think that covers it, but I also noticed that you avoid `expand_vectors` for all but the first iteration. I did not include this yet, because I was not sure why you included this. Is it a performance thing?

PeterRabbit95:
Hey thanks a lot for fixing this! Unfortunately, I do not have time to properly look into it tonight but I'll do it asap. I just gave a quick look and I like how you organized the code. 

This said, as for the `expand_vectors` I excluded it for two reasons:
- The test was failing and I could not figure out why so I thought it was safer to remove it (also I never used it so I do not have a good understanding of it)
- I assumed once you expand all the vectors there is no need to do it again (as I said, I am not sure if this is actually the case)

PeterRabbit95:
Hi Jack, I just tested your solution but unfortunately it does not fully work on my model. Unfortunately, the test that I previously implemented was not enough to highlight the problem. 


I was not allowed to push on [https://github.com/pymoca/pymoca/commits/issue-214]( https://github.com/pymoca/pymoca/commits/issue-214) so I made a few changes and pushed them on my fork: 
[https://github.com/PeterRabbit95/pymoca/tree/PeterRabbit95-issue-214](https://github.com/PeterRabbit95/pymoca/tree/PeterRabbit95-issue-214).


Basically I cleaned my big model from all the confidential stuff and used it as the test model for iterative simplification on commit 1cb64f49b1488ce5ba9029044297f3c2423f50f1 . As I said the test fails and it is because when we drop a model variable we do not delete it from the aliases which means that on the second call, when attributes from aliases are merged to canonical variables, the variable is not found in all_states and a key error is raised. 

In commit 8bad233bb32047a825e39acc096c62ad12284465 I implemented again the same solution that keeps track of substituted aliases in `alias_reliation` as I did in commit f0415d5b9a0fe57307d07118727048528bf58bda.

With this last change, everything works as long as we also keep `replace_constant_values` to `True`. If we deactivate it, two problems arise:
- The first is not necessarily connected to the `iterative_simplification`, so I opened another issue #218 
- The second is that, accordingly to how we solve #218, the logic you added (commit: f0f025c182447e1018d48bcd4d98f8dca85c4ce7) in `replace_constant_values` to eliminate the constants with aliases should maybe be moved to `eliminate_constant_assignments`.

jackvreeken:
Added two commits on top of issue-214 I had. I took a different approach, because I do not want to have information about what aliases were already substituted in a previous pass in the AliasRelation class if we can avoid it. This one seems to work for the larger model you provided.

I also added you with "write" rights to the organization, which should mean you can now create branches.

PeterRabbit95:
Your solutions looks very good! I pushed something more to the branch in order to solve a naming problem that prevented the test to go through and I made an explicit call to .simplify(...). You can merge to the master if you believe.

jackvreeken:
Alright, merge the issue-214 branch.

PR #220, opened 2020-03-18
CasADi: Keep track of simplified variables when eliminated

PeterRabbit95:
Whenever a variable is eliminated it is not deleted but it gets moved to a list in a new field of `model` object called `simplified_variables`. This opens different possibilities, such as accessing results for all the variables of a model even though they do not take part in the simulation.


pietropelizzari:
I will start to work on this change again soon. For the time being, I am setting this PR to draft as it _clearly_ not ready to be reviewed and merged.

PR #221, opened 2020-03-18
CasADi: simplify equations with additions and subtractions

PeterRabbit95:
Add a logic to simplify equations such as `(-3*x) - y` to `3*x + y`


test/gen_casadi_test.py
```diff
@@ -2420,6 +2426,36 @@ def test_iterative_simplification_large(self):
...
+        ref_model = Model()
+
+        y = ca.MX.sym('y')
+        x = ca.MX.sym('x')
+        z = ca.MX.sym('z')
+
+        ref_model.alg_states = [Variable(y), Variable(x), Variable(z)]
+        ref_model.equations = [x + y, z, y]
 
+        self.assert_model_equations_equivalent(ref_model, casadi_model)
```
jackvreeken: Why is `assert_model_equivalent` not good enough? Also, please keep two blank lines before the `if __name___` part :)

src/pymoca/backends/casadi/model.py
```diff
@@ -728,22 +728,35 @@ def get_derivative(expr):
...
-                # These are binary ops and can get a little tricky
-                # For now, we just drop constant divisors or multipliers
-                elif eq.n_dep() == 2 and eq.op() in binary_ops:
-                    if eq.dep(1).is_constant():
-                        return factor_and_simplify(eq.dep(0))
-                    elif eq.dep(0).is_constant() and eq.op() == ca.OP_MUL:
-                        return factor_and_simplify(eq.dep(1))
-
+                elif eq.n_dep() == 2:
+                    # Drop constant divisors or multipliers
```
jackvreeken: Not your change, but I'm thinking this simplification is actually rather dangerous. For scaling of equations we actually might want to allow dividing an equation by a constant... I always feel rather unsure about solely leaving it to a solver like IPOPT to scale constraints.

jackvreeken:
Aside from those two comments, it looks good to me (especially with the comment giving the example, and the test!)

src/pymoca/backends/casadi/model.py
PeterRabbit95: Well, the elimination of constant dividers was there before (together with constant multipliers). I am only adding the additions and subtractions binary operations that can be merged with the negative unary operations. 

The point you raise about IPOPT is definitely true. In any case, I am not sure that scaling a variable directly in the Modelica model would be the best way to go. If you were to tweak the Modelica model with the solution process already in mind (be it an optimization or a simple simulation) then you could simply leave the simplification options deactivated.

test/gen_casadi_test.py
PeterRabbit95: Sorry about the black lines eheh :) with pycharm I have no excuses :)) I'll fix it!

As for `assert_model_equivalent` I think it is not enough because only the variables are compared (we could maybe change the name because it could be misleading).

I could definitely use `assert_model_equivalent_numeric` but in this case also the order of the equations is compared.

What do you think?

pietropelizzari:
I will start to work on this change again soon. For the time being, I am setting this PR to draft as it _clearly_ not ready to be reviewed and merged.

PR #222, opened 2020-03-24
CasADi: bugfix on iterative detect alias

PeterRabbit95:
The previous logic would fail in the following situation:
- in iteration 1: A is aliased to B and A is deleted
- in iteration 2: B is aliased to C, the canonical variable of A is then C. 
The check `alias in old_alias_relation.aliases(canonical)` which in these case is `A in old_alias_relation.aliases(C)` returns False and an error is raised when A is not found in all_states

I was not able to create a simple example to showcase this. I have this problem with a rather big model. In any case, this solution works with the previous tests and it is rather similar from a logical point of view.

### DIFF
--- a/src/pymoca/backends/casadi/model.py
+++ b/src/pymoca/backends/casadi/model.py
@@ -888,7 +888,8 @@ class Model:
                 fixed = canonical_state.fixed
 
                 for alias in aliases:
-                    if alias in old_alias_relation.aliases(canonical):
+                    if len(old_alias_relation.aliases(alias)) > 1 and \
+                            alias not in old_alias_relation.canonical_variables:
                         # We already handled this alias in a previous pass of `detect_aliases`
                         continue
 
--- a/test/gen_casadi_test.py
+++ b/test/gen_casadi_test.py
@@ -1436,6 +1436,43 @@ class GenCasadiTest(unittest.TestCase):
         self.assertSetEqual(p.aliases, {'z'})
         self.assertEqual(len(casadi_model.alg_states), 0)
 
+    def test_deleted_canonical_variable(self):
+        txt = """
+            model DeletedCanonicalVariable
+              Real A(min=1.0, max=4.0);
+              Real B(min=2.0, max=5.0, nominal=10.0);
+              Real x(min=3.0, max=6.0);
+              constant Real c = 0;
+            equation
+              der(x) = 1;
+              A = B;
+              x = B + c;
+            end DeletedCanonicalVariable;
+        """
+        ast_tree = parser.parse(txt)
+        casadi_model = gen_casadi.generate(ast_tree, 'DeletedCanonicalVariable')
+
+        # A is aliased to B and A is deleted
+        casadi_model.simplify({'detect_aliases': True})
+
+        # B is aliased to x, the canonical variable of A is then x
+        casadi_model.simplify({'replace_constant_values': True})
+        casadi_model.simplify({'detect_aliases': True})
+
+        ref_model = Model()
+
+        x = ca.MX.sym('x')
+        der_x = ca.MX.sym('der(x)')
+
+        ref_model.states = list(map(Variable, [x]))
+        ref_model.states[0].min = 3.0
+        ref_model.states[0].max = 4.0
+        ref_model.states[0].nominal = 10.0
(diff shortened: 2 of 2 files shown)


jackvreeken:
I'll see if I can figure out what is wrong, and whip up a test case. I would like to avoid fixing bugs without a test case, as I have cut myself hard in the past a few times that way.

PeterRabbit95:
Hi, thanks for your feedback! I think I can try again to create a test :) I have a good understanding of the problem so with some more effort I should be able to

PeterRabbit95:
I implemented a test that does exactly what I described in the PR. Two things to keep in mind:

- I did not use iterative_simplification but I manually combined multiple calls to .simplify()
- I am not sure how to properly write the assertion with Unittest. The test already works properly because a KeyError is raised when the test fails but I did not use any specific unittest method. I am not sure if this is a problem.

jackvreeken:
That fix looks perfect! And using multiple calls to simplify() in the test is fine by me. 

I would suggest to use to put the model in a string instead of a separate file. The reason is that a call to `transfer_model` will parse the _entire_ folder, which means that all .mo files for all tests are parsed (even though they are not relevant). This is rather slow as you can imagine. Please also put the test near the other alias tests, e.g. just below the parameter_aliases test (a bug you found before):

```Python
    def test_parameter_aliases(self):
        [...]
        self.assertEqual(len(casadi_model.alg_states), 0)

    def test_deleted_canonical_variable(self):
        txt = """
            model DeletedCanonicalVariable
              Real A;
              Real B;
              Real x;
              constant Real c = 0;
            equation
              der(x) = 1;
              A = B;
              x = B + c;
            end DeletedCanonicalVariable;
        """
        ast_tree = parser.parse(txt)
        casadi_model = gen_casadi.generate(ast_tree, 'DeletedCanonicalVariable')

        # A is aliased to B and A is deleted
        casadi_model.simplify({'detect_aliases': True})

        # B is aliased to x, the canonical variable of A is then x
        casadi_model.simplify({'replace_constant_values': True})
        casadi_model.simplify({'detect_aliases': True})

    def test_simplify_expand_vectors(self):
        # Create model, cache it, and load the cache
        compiler_options = \
        [...]
```

As far as a unit test is concerned: I would say that in this case you want to be sure that all relevant attributes (e.g. a `min` attribute) is actually the minimum of that of `A`, `B`, and `x`. And that the max is the max of all three, no matter the order in which they are processed. That way we can be sure that all the copying of attributes is done correctly, and we do not accidentally skip the processing of an alias. So if you have a model like:

```Modelica
            model DeletedCanonicalVariable
              Real A(min=1.0, max=4.0);
              Real B(min=2.0, max=5.0, nominal=10.0);
              Real x(min=3.0, max=6.0);
              constant Real c = 0;
            equation
              der(x) = 1;
              A = B;
              x = B + c;
            end DeletedCanonicalVariable;
```

you can put something like the following code at the end of the test, that should cover it:

```Python
       [...]

        # B is aliased to x, the canonical variable of A is then x
        casadi_model.simplify({'replace_constant_values': True})
        casadi_model.simplify({'detect_aliases': True})

        ref_model = Model()

        x = ca.MX.sym('x')
        der_x = ca.MX.sym('der(x)')

        ref_model.states = list(map(Variable, [x]))
        ref_model.states[0].min = 3.0
        ref_model.states[0].max = 4.0
        ref_model.states[0].nominal = 10.0
        ref_model.der_states = list(map(Variable, [der_x]))
        ref_model.equations = [der_x - 1]

        self.assert_model_equivalent_numeric(casadi_model, ref_model)
```

Simply put, `A` has the governing maximum (the lowest one of the three), `x` has the governing minimum (the highest one of the three), and `B` has the governing nominal (only one that specifies it). We make sure that this is actually the case for the one state we end up with.

PeterRabbit95:
Thanks a lot for the feedback! I implemented what you suggested (basically copy-paste :)). How does it work in this cases? Should you should give a final "green light" or I can already squash and merge?

jackvreeken:
I would say you squash the fixes into one commit, and then you can just go ahead and merge. For bigger pull requests (with multiple commits, touching many files, etc) I might want to take another look after a final squash/cleanup, or would ask you to another look on my own work.

PR #223, opened 2020-04-02
Parse Optimica (Modelica extension for optimal control)

PeterRabbit95:
I have been working for a while on this new functionality and now that I have a first working version I'd like to share it with you in order to get some impressions. Optimica is a Modelica extension proposed in this [paper](https://www.researchgate.net/publication/253089188_Optimica-An_Extension_of_Modelica_Supporting_Dynamic_Optimization) that embeds the information needed for dynamic optimization directly in Modelica code. The main features are:
- a new class type called optimization
- new class attributes (objective function, initial and final time)
- accessing a variable at a specific time (in order to define the cost function and the point constraints)

As far as I know, Optimica is only compatible with Jmodelica. There is already a big collection of models available written in Optimica.

Adding this feature would require changes all over the workflow of pymoca, from the ANLTR parser to the CasADi generator walker (or the other backends).

I already implemented (and committed) a first working version. I focused only on the CasADi backend. The following model is correctly parsed:

```Modelica
optimization VDPocp(objective = x1(finalTime)^2 + x2(finalTime)^2, objectiveIntegral = x1^2 + u^2 + x2^2, startTime = 0, finalTime = 0)
    Real x1(start = 0);
    Real x2(start = 1);
    input Real u(min = -0.75, max = 1, free = true, initialGuess = 0);

    equation
    der(x1) = (1 - x2^2) * x1 - x2 + u;
    der(x2) = x1;

    constraint
    x2(4) >= 1;
end VDPocp;
```

The procedure to import the model is the standard one:

```python
from pymoca.backends.casadi.api import transfer_model
model = transfer_model('model_path', 'VDPocp')
print(model)
```

New attributes of model have been created to accommodate the new information:

```
Model
time: time
states: [x1[1,1]:float, x2[1,1]:float]
der_states: [der(x1)[1,1]:float, der(x2)[1,1]:float]
alg_states: []
inputs: [u[1,1]:float]
outputs: []
constants: []
parameters: []
equations: [MX((der(x1)-((((1-sq(x2))*x1)-x2)+u))), MX((der(x2)-x1))]
initial equations: []
constraints: [MX((x2(4)-1))]
optimization_attributes: {'objective': MX((sq(x1(finalTime))+sq(x2(finalTime)))), 'objectiveIntegral': MX(((sq(x1)+sq(u))+sq(x2))), 'startTime': 0, 'finalTime': 0}
```


jackvreeken:
I have no issue at all with adding this, seems useful! As long as we do not break the old API it's fine by me. There is sort of a requirement that Optimica is a superset of Modelica I guess, but I do not know if that's the case. If behavior for a certain keyword differs between Optimica and Modelica, it will become rather iffy.

I'm not sure how strict we want to be on allowing e.g. the "constraints" keyword; inequalities might actually be useful for normal `class`es as well.

pietropelizzari:
It is good to know that you would be open to including Optimica. :) 

I think we can safely state that Optimica is a superset of Modelica. Something important to say is that there is not a standard unique definition of Optimica (e.g the behavior described in the paper I linked and the behavior implemented in Jmodelica is sometimes slightly different). This could be an advantage because, as you said, we could change a keyword such `constraints` to `inequalities`. On the other hand, it would be good to keep it compatible with the example collection of Jmodelica.

I think that inequalities in simple simulations could be represented by the assert statement, which I, as far as I know, is not available in pymoca.

I totally agree with the simplification topic. I think a good starting point would be `_substitute_metadata`. I will not have time to work on it in the next few weeks but afterward, I could start making it robust and writing some tests. 

Another specific thing I want to change is the `optimization_attributes` field in `Model`. As for now, it is a dictionary that could potentially accept every entry. Of course, this does not make sense as we have a list of acceptable keys.

pietropelizzari:
I have not worked on this for a long time but I will soon update to proposal including merging the changes which were added to the parser in the mean time. I am converting it to draft for the time being.

PR #225, opened 2020-08-03
CasADi: Support string parameters in model

jackvreeken:
- [x] Do we want the dae_/initial_dae_/variable_metadata CasADi functions to be a function of string parameters as well? Currently that is the case, but it can look somewhat weird.

Some options are:
1. Keep as is; `model.parameters` contains all parameters regardless of type.  CasADi functions are a function of all parameters.
2. `model.parameters` contains all parameters regardless of type.  CasADi functions are a function of only the numeric parameters.
3. Add `model.string_parameters`, as a separate parameter group for non-numeric parameters. Not an input to any of the CasADi functions.

EDIT:
Options 3 is probably the cleanest one, so going with that one.

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -36,6 +36,8 @@ class CachedModel(Model):
         self.outputs = []
         self.constants = []
         self.parameters = []
+        self.string_constants = []
+        self.string_parameters = []
         self.time = ca.MX.sym('time')
         self.delay_states = []
         self.delay_arguments = []
@@ -57,6 +59,8 @@ class CachedModel(Model):
         r += "outputs: " + str(self.outputs) + "\n"
         r += "constants: " + str(self.constants) + "\n"
         r += "parameters: " + str(self.parameters) + "\n"
+        r += "string_constants: " + str(self.string_constants) + "\n"
+        r += "string_parameters: " + str(self.string_parameters) + "\n"
         return r
 
     @property
@@ -212,6 +216,8 @@ def save_model(model_folder: str, model_name: str, model: Model,
         # Describe variables per category
         for key in ['states', 'der_states', 'alg_states', 'inputs', 'parameters', 'constants']:
             db[key] = [e.to_dict() for e in getattr(model, key)]
+        db['string_constants'] = model.string_constants
+        db['string_parameters'] = model.string_parameters
 
 
         # Caching using CasADi functions will lead to constants seemingly
@@ -348,6 +354,8 @@ def load_model(model_folder: str, model_name: str, compiler_options: Dict[str, s
         model.outputs = db['outputs']
         model.delay_states = db['delay_states']
         model.alias_relation = db['alias_relation']
+        model.string_constants = db['string_constants']
+        model.string_parameters = db['string_parameters']
 
         # Evaluate variable metadata:
         parameter_vector = ca.veccat(*[v.symbol for v in model.parameters])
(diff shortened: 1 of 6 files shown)


jackvreeken:
@pietropelizzari If you have time, could you give this a quick check? Basically adding a `string_parameters` property to a model. String parameters do not appear in the DAE, so simplifications etc are fully bypassed this way.

pietropelizzari:
@jackvreeken sure I can look at it. Sorry for the late answer but it did not have access to my laptop in the last couple of days. I'll check it within today

jackvreeken:
@pietropelizzari Cool! Let me know what you find. I'll squash the fixup commits (current and future ones) when you're done reviewing and I get your OK.

pietropelizzari:
Sorry for the late answer. Anyway, I just looked through your changes. I honestly struggle a bit in understanding the role of String Parameters in Modelica. Right now, I do not see how a string parameter can ever take part in an equation so excluding them from the CaADi functions seems reasonable. 


Generally, I would probably go for the second approach you described. I feel like creating a dedicated "string_parameters" attribute does not "match" with the Modelica grammar structure. Anyway, I do not think it makes a big difference.


I made two commits. The first one is pretty self-explanatory and I made just some minor changes.


In the second commit, I created 2 very simple tests to show an error that rises with String constants. As we do not have an additional field "string_constants" in the model, when we declare a String constant a StringVariable object is added to model.constants and this can create conflicts. I am not sure what is the best way to solve this issue. I think we could either: 
- go for the second approach you described and ignore both string constants and parameters when creating CasADi functions,
- create an additional string_constants attribute,
- merge String constants and parameters in a single attribute.


Furthermore, I notice that, as stated in the [Modelica Grammar specification](https://specification.modelica.org/master/class-predefined-types-and-declarations.html#string-type), the default value for _fixed_ should be _True_ for constants and parameters of String type. I am afraid this might be tricky to solve as I think we should act on the Symbol class defined in ast.py

https://github.com/pymoca/pymoca/blob/1fdee00fe4f5a8e61a540a7fb2ecb4db4344e1e7/src/pymoca/ast.py#L303-L330


One last thing is that I was not able to check the behavior of the caching/loading for string_parameters. I have seen you handled them differently wrt regular parameters but I have never used caching and I have no clue what exactly happens. https://github.com/pymoca/pymoca/blob/1fdee00fe4f5a8e61a540a7fb2ecb4db4344e1e7/src/pymoca/backends/casadi/api.py#L214-L218

jackvreeken:
1. As far as the caching is concerned, there is a test ("test_cache_metadata" iirc) that checks whether caching works as intended.
2. The default value for fixed True/False I agree with you on, but that is something that I feel we should pick up later as it's more on the parser/ast level.
3. The reason I went for the string_parameters route, is that I feel it's rather weird to have a CasADi DAE function that is a function of `[states, alg_states, parameters, ...]`, but then somehow always have it be the case that it is only a function of a subset of those. Furthermore, if you make it a function of all parameters, you have to provide some (numeric) value for these string parameters. You could say "just use NaN", but that's rather ugly imo. The other option is to make the DAE function have inputs like ```[states, alg_states, *[p for p in parameters if isinstance(p, Variable), ...]```, but then this calling signature is not entirely obvious for applications using Pymoca.

The Pymoca API is still unstable (and we'll keep it that way for now), so even if we settle for `string_parameters` and `string_constants` with this MR, we can still change it in the future if new insights/experiences point to a different solution.

pietropelizzari:
Then I'd say we can add "string_constant" and merge 👍. Maybe it is worth to delete the tests I created to show the constant problem and create a String constant in one of the other tests.

PR #241, opened 2021-02-17
Modelica: Add __repr__ to AST Nodes

kentrutan:
Add `__repr__` to all AST `Node`'s to improve debugging.

In the process I also changed some `__str__` methods. This could break any code that depended on the prior `str()` format. In particular, the `__str__` method of `Node` itself (that is inherited by many subclasses) now tries to trim out trivial/empty attributes to make a more readable json printout. This "feature" could be reverted if necessary.

The the `ComponentRef` class method `concatenate` is changed to a regular method and now takes a single `ComponentRef` instead of a list. This does not appear to be used in any other pymoca code, but again could break other code using it. I do use it in the added AST tests, but the main reason for changing this is efficiency for an upcoming `import` addition. This could be easily reverted if necessary and/or moved to the upcoming `import` pull request.

Finally, we would like to update the AUTHORS and LICENSE files. Please propose any additions needed to the AUTHORS file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,3 +6,6 @@ James Goppert
 The CasADi backend was funded by Deltares and written by:
 Joris Gillis (Yacoda)
 Jorn Baayen (Deltares)
+
+Contributions to Modelica frontend by:
+Kent Rutan (Caterpillar Inc.)
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2016, James Goppert
+Copyright (c) 2016-2021, James Goppert and the Pymoca contributors.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -2,7 +2,7 @@
 """
 Modelica AST definitions
 """
-from __future__ import print_function, absolute_import, division, print_function, unicode_literals
+from __future__ import print_function, absolute_import, division, unicode_literals
 
 import copy
 import json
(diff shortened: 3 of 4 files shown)


jackvreeken:
So if understand you correctly, there'd eventually be three commits for this PR:
1. Add currently missing __repr__ for many classes (very useful indeed!)
2. Add and improve __str__ (separate in case anyone actually depended on it, although I doubt it).
3. ComponentRef concatenate changes/repurposing

As far as (3) is concerned, I think the `.concatenate` is a remnant of the times before 514402d138110ec44ea9ba3bc126328337a3933e (i.e. before we flattened in a way similar to the Modelica specification). I guess I just forgot to remove it in that commit.

Going over the changes, they looks good to me. Could you clean up the PR a bit so I can merge it? Something like:
- split your changes in these three commits
- add some explanation in every commit on why the change is useful/necessary. You can just copy paste this from your opening comment in this PR.
- for the commit on the ComponentRef, it would be nice if you could add to the commit message that the .concatenate method was made superfluous by 514402d1 (and hasn't been relied upon since), and that we're now repurposing it.
- rebase on top of master. Pymoca uses the "linear history" convention when merging to master (i.e. no merge commits). For WIPs I don't care what you do of course ;)

I rarely ever change the authors file (I'm not even in there myself), but with company affiliations I understand why you'd want to do this. I'm fine with you adding that information as a first separate commit, or make it part of your first commit along with e.g. the __repr__ changes.

kentrutan:
Thanks for your quick review and guidance. Totally makes sense. I was heading in that direction as I wrote the PR comment, but was eager to get the first one submitted! Should I have put WIP in the title? Still learning how you are doing things.

I appreciate you accepting changes to the legal stuff. You should definitely be in the AUTHORS file! Please email me to discuss.

jackvreeken:
Yeh, no worries about all that! Better to just get the discussion started on all this (the contents, and the process).

I generally do put "WIP: ..." in the title when something is not ready for merge yet (but I want to have feedback on it first).

kentrutan:
I pulled out the other stuff so this is more of a pure `__repr__` add with hopefully no backward compatibility issues. I will include the other changes in separate PRs.

@jackvreeken  please add your name and any others you think should be added to the AUTHORS file.

PR #243, opened 2022-02-18
Add compiler tool options

kentrutan:
Update the compiler tool and add options to support the SymPy backend in addition to CasADi in the previous version. Hopefully this will make it easier to use Pymoca without having to resort to python scripting. I know the SymPy backend is not being actively developed, but I wanted to support at least two for starters. The xml backend could be easily added if that would be useful to anyone. For more info, see the comment on 38d44c2

Since I do not have experience with CasADi, it would be great of someone who does could try this out in an actual workflow.

The longer-term intention is to make it easier to add backends to the compiler tool and incorporate more options to support planned improvements to parse Modelica libraries.

This pull request also includes a commit with new notebook that I used for testing the SymPy output from the test case model Spring.mo. It is a copy/paste/modify of the Spring.ipynb.

PR #244, opened 2022-02-28
Don't build AST if parse fails

kentrutan:
Before this change, `parser.parse()` would continue on its merry
way trying to build an AST when there were syntax errors from ANTLR.
At the least it is wasted time.

Also fixed type errors found by mypy and in the process found that
there was a bit of duplicate code between `ASTListener` members
`ast_result` and `file_node`.

### DIFF
--- a/src/pymoca/parser.py
+++ b/src/pymoca/parser.py
@@ -6,7 +6,7 @@ from __future__ import print_function, absolute_import, division, unicode_litera
 
 import antlr4
 import antlr4.Parser
-from typing import Dict
+from typing import Dict, List, Union
 from collections import deque, OrderedDict
 import copy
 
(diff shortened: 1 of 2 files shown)


jackvreeken:
Is this also catching errors with missing semicolons and the like?

kentrutan:
Yes it should catch syntax errors flagged by ANTLR and gracefully return before trying to build the AST. 

I have not extensively dug into how it crashed and burned before. Let me know if you want me to investigate if it fixes specific things that have been a pain in the past. It was a pretty old commit that last touched that code.

kentrutan:
I changed the commit message, otherwise the same contents. Please go ahead and merge this into master if you think it is OK. I realized today that #243 depends on this change, so if you could accept this pull request first, then I'll rebase #243 on it and push some updates.

As you can tell, I'm still getting up to speed on git and github as well as how you are managing this repo, so please keep teaching me how to fish. Thanks!

jackvreeken:
No worries, it's awesome that you're taking the time to polish things!

For commit messages, I'd suggest just looking in the commit log, and copy that style (which I copied from e.g. the Linux kernel/x264/VLC/ffmpeg/...).

Issue #4, opened 2016-04-09
Support modelica standard library


jgoppert:

mjfwest:
Is there anything I can do to help on this issue? I am new to Pymola, but I would really like this enhancement. Just let me know where to start.

jackvreeken:
I'm not entirely sure what is needed, but one way to start would be to either
1. Make sure the entire library is parsed correctly. This is a big undertaking.
2. Only parse files that are necessary. Currently all files in a library folder are parsed recursively, which means a _lot_ of parsing for that one file you need.
3. Strip down the Modelica library to that one or two components you need first, and check if you can get a simple model with those components to work.

kentrutan:
I have implemented the `import` statement while working at Caterpillar Inc. I have official Caterpillar approval to open source my changes, so am working to get the code ready to submit a pull request. Looking forward to working with this community!

jackvreeken:
Nice, that must have been some hellish work! Scoping is a tough one I've burned myself on (and got very frustrated with Modelica) many times. I look forward to seeing how you solved it :)

kentrutan:
It was a challenge! It's a first pass and I'm sure working together we can come up with a better solution, so looking forward to some pull request discussion.

kentrutan:
Apologize for all the toggling of the status flag above. Didn't realize it was tracking my every move haha.

As noted above the scope of this issue is huge, so I'll look to create some smaller scope issues to add to the Modelica Compiler project. Most of what I'll be doing near term is related to this.

PR #246, opened 2022-03-10
Modelica: Add import statement

kentrutan:
This adds support for the Modelica `import` statement. It implements the name lookup logic from the Modelica Language Spec chapters 5 and 13 which is more succinctly outlined in the [Modelica by Example book](https://mbe.modelica.university/components/packages/lookup/).

This is a step toward using existing Modelica libraries per issue #4. It reduces the number of parse errors from 105 to 89 out of a total of 216 files in the Modelica Standard Library version 3.2.3. For MSL 4.0.0 parse errors are reduced from 586 to 278 out of a total of 2512 files. A major shortcoming, however is that all dependencies must be parsed in order for name lookup to succeed. Implementing support for `MODELICAPATH` and implementing lazy parsing per the [README TODO list](https://github.com/pymoca/pymoca#readme) will be quite useful as a followup to this PR.

Following this PR, it will be relatively easy to fix the SIunits workaround [mentioned by @jackvreeken in issue #4](https://github.com/pymoca/pymoca/issues/4#issuecomment-348270333).

The [Modelica-Compliance library](https://github.com/modelica/Modelica-Compliance) is used in the new test cases and I believe it will be used enough in future test cases that it makes sense to add as a submodule.


jackvreeken:
Sorry, I missed this one a few weeks back. Will have a look later this week, but from the commit messages / PR message it already looks like you did some amazing work!

kentrutan:
>  from the commit messages / PR message it already looks like you did some amazing work!

Well you may not find it so amazing when you look at the code! I implemented it a year ago and as I look at it now I want to improve it and test it more, but this seems like a good milestone for now that has worked for a few test cases.

kentrutan:
With the commit I just added I think this is ready to merge.

PR #257, opened 2022-12-08
DELME: Just running some Github Actions

jackvreeken:


kentrutan:
Probably should update the version in create_conda_env.sh as well. I tried to push once, but you are rebasing often so I'll let you do it. Appreciate you fixing the CI/CD.

jackvreeken:
Well spotted! Will update the create_conda one as well

PR #263, opened 2023-03-08
Modelica: Add `+` expression prefix

kentrutan:
This allows expressions like `(+2)` per the language standard. This change eliminates parse errors for 45 files (about 2%) of the Modelica Standard Library 4.0.0 in the Magnetic, Electrical, Mechanics, Fluid, and Clocked domains.

Tests are added for both + and - prefixes.


jackvreeken:
LGTM. Are you planning on adding backend support as well? Also, do I understand correctly that the "Modelica: " prefix means "Grammar/Parser: " (i.e. you're working on fixing all kinds of parsing/grammar issues)?

Did a quick check, and the SymPy backend already seems to support the + operator (with one operand). The CasADi one doesn't yet, but easy to add with a block like (in `casadi\generator.py`, in `exitExpression`):

```Python
        elif op == '+' and n_operands == 1:
            src = self.get_mx(tree.operands[0])
```

kentrutan:
> Are you planning on adding backend support as well? Also, do I understand correctly that the "Modelica: " prefix means "Grammar/Parser/AST: , i.e. not touching backends?

"Modelica:" refers to the [Modelica Compiler
](https://github.com/pymoca/pymoca/projects/1#card-88412019) project to support the Modelica Standard Library. I take that to mean on the frontend with the goal to close out #4. If the tests show a backend is broken by a frontend change, I will try to fix. My CasADi and SymPy skills are pretty minimal, so I can use some help implementing support for frontend changes into the backends.

> 
> Did a quick check, and the SymPy backend already seems to support the + operator (with one operand). The CasADi one doesn't yet, but easy to add with a block like (in `casadi\generator.py`, in `exitExpression`):
> 
> ```python
>         elif op == '+' and n_operands == 1:
>             src = self.get_mx(tree.operands[0])
> ```

Glad you checked! My primary goal currently is to get this repo up to speed with all the frontend improvements I have made on my fork, while also cleaning them up, testing, and documenting. All that means I have little time for backend improvements for now.

jackvreeken:
Ah, support for Modelica Standard Library is indeed a big/ambitious one. Don't worry about backend stuff then, I'll try to handle the CasADi support/tests/etc for new features where I can :)

PR #264, opened 2023-03-13
CasADi: Support + expression prefix

jackvreeken:
Parser support for this was added in 81b4a51 .

Note that a negative prefix was already supported by the CasADi backend.


jackvreeken:
And there is your Modelica test case again @kentrutan, but now as part of the CasADi backend tests ;)

kentrutan:
I took a look and tested on my machine. Looks good to go!

PR #265, opened 2023-03-13
Tree: Fix scoping of inherited symbols

kentrutan:
This fixes cases where inherited symbol type lookup fails because the scope of the type was lost during instantiation.  One example where it fails is in flattening MSL 4 `Modelica.Electrical.Analog.Basic.Resistor` where the symbol type name `PositivePin` is referenced in an inheritance hierarchy two levels deep.

See Modelica Language Spec 3.5 section 5.6.1.2 and related sections.

This implementation adds a parent attribute to the `ast.Symbol` class that is set during creation of the AST. If local lookup fails, the original symbol parent is then used as the starting point for symbol type lookup in `tree.build_instance_tree()`.

A test case is added that is similar to MSL 4 Resistor that failed.

### DIFF
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -399,6 +399,7 @@ class Symbol(Node):
         self.order = 0  # type: int
         self.visibility = Visibility.PRIVATE  # type: Visibility
         self.class_modification = None  # type: ClassModification
+        self.parent = None # type: Class
         super().__init__(**kwargs)
 
     def __str__(self):
--- a/src/pymoca/parser.py
+++ b/src/pymoca/parser.py
@@ -637,14 +637,14 @@ class ASTListener(ModelicaListener):
             s.type = copy.deepcopy(clause.type)
 
     def enterComponent_declaration(self, ctx: ModelicaParser.Component_declarationContext):
-        sym = ast.Symbol(order=self.sym_count)
+        sym = ast.Symbol(order=self.sym_count, parent=self.class_node)
         self.sym_count += 1
         self.ast[ctx] = sym
         self.symbol_node = sym
         self.comp_clause.symbol_list += [sym]
 
     def enterComponent_declaration1(self, ctx: ModelicaParser.Component_declaration1Context):
-        sym = ast.Symbol(order=self.sym_count)
+        sym = ast.Symbol(order=self.sym_count, parent=self.class_node)
         self.sym_count += 1
         self.ast[ctx] = sym
         self.symbol_node = sym
@@ -654,7 +654,7 @@ class ASTListener(ModelicaListener):
         if self.symbol_node is not None:
             self.ast[ctx] = self.symbol_node
         else:
-            sym = ast.Symbol(order=self.sym_count)
+            sym = ast.Symbol(order=self.sym_count, parent=self.class_node)
             self.sym_count += 1
             self.ast[ctx] = sym
             self.symbol_node = sym
--- a/src/pymoca/tree.py
+++ b/src/pymoca/tree.py
@@ -161,7 +161,7 @@ class TreeWalker:
         endless recursion by skipping references to e.g. parent nodes.
         :return: True if child needs to be skipped, False otherwise.
         """
-        if isinstance(tree, ast.Class) and child_name == 'parent' or \
+        if isinstance(tree, (ast.Class, ast.Symbol)) and child_name == 'parent' or \
                 isinstance(tree, ast.ClassModificationArgument) and child_name in ('scope', '__deepcopy__'):
             return True
         return False
@@ -366,7 +366,11 @@ def build_instance_tree(orig_class: Union[ast.Class, ast.InstanceClass], modific
 
         try:
             if not isinstance(sym.type, ast.InstanceClass):
-                c = extended_orig_class.find_class(sym.type)
+                try:
+                    c = extended_orig_class.find_class(sym.type)
+                except ast.ClassNotFoundError:
+                    # If not found in local scope, search in original scope
+                    c = sym.parent.find_class(sym.type)
             else:
                 c = sym.type
         except ast.FoundElementaryClassError:
--- a/test/models/InheritanceInstantiationResistor.mo
+++ b/test/models/InheritanceInstantiationResistor.mo
@@ -1,0 +1,27 @@
+// Example to test inherited name lookup with a case similar to MSL v4
+// Modelica.Electrical.Analog.Basic.Resistor where the symbol type name
+// PositivePin is referenced in an inheritance hierarchy two levels deep.
+// It also tests modifying something other than value.
+package P
+  package T
+    class Voltage = Real(nominal=1);
+  end T;
+  package I
+    class Level1
+      // i.e. OnePort
+      extends Level2(p.v(nominal=10), n.v(nominal=10));
+    end Level1;
+    class Level2
+      // i.e. TwoPin
+      Level3 p(v(nominal=20)), n(v(nominal=20));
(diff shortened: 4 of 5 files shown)


jackvreeken:
Aside from the missing newline at the end of the .mo file, looks good.

Related question, do you cross-check the flattening of these models against OpenModelica? I can't get OMC to play along, it complains about the Real inheritance.

I was wondering what the inheritance should be if I add a Level3 at the end:

```Modelica
package P
  ... unchanged ...
    class Level3 // Overrided max attribute
      // i.e. PostitivePin
      T.Voltage v(nominal=30, max=10.0);
    end Level3;
  end I;
  model M // Unchanged
    // i.e. Resistor
    extends I.Level1(p.v(nominal=0), n.v(nominal=0));
  end M;

  class Level3  // NEW, with a different max attribute
    T.Voltage v(nominal=30, max=20.0);
  end Level3;
end P;
```

The resolution should then prefer which of the two? I.e. 

```Python
        self.assertEqual(flat_tree.classes['P.M'].symbols['p.v'].max.value, 20.0)
        self.assertEqual(flat_tree.classes['P.M'].symbols['n.v'].max.value, 20.0)
```

or 

```Python
        self.assertEqual(flat_tree.classes['P.M'].symbols['p.v'].max.value, 30.0)
        self.assertEqual(flat_tree.classes['P.M'].symbols['n.v'].max.value, 30.0)
```

kentrutan:
I do check against omc. It gives the following for my original test case:
```
% omc --version
OMCompiler v1.18.0
% omc -i=P.M test/models/InheritanceInstantiationResistor.mo
class P.M
  Real p.v(nominal = 0.0);
  Real n.v(nominal = 0.0);
end P.M;
```
For your changes it gives:
```
% omc -i=P.M test/models/InheritanceInstantiationResistor.mo
class P.M
  Real p.v(max = 10.0, nominal = 0.0);
  Real n.v(max = 10.0, nominal = 0.0);
end P.M;
```
Your additions show an issue with my current commit. Very perceptive! I'll take a look.

BTW, I set the option in my editor to always put a newline in, so shouldn't have that problem going forward. Another good catch!

I very much appreciate having someone else looking at my code! Thanks!

jackvreeken:
Ah, I totally forgot about the "-i" option. Upgraded to the latest version of OpenModelica in the process; might have been a bit of both.

Glad to be of help! And this name resolution / finding class is something I repeatedly ran into issues with, so that's why I thought of it. You probably already figured this out by looking at forum discussions, but Dymola and OpenModelica don't/didn't always agree, although it certainly has gotten a lot better over time with all the test cases.

Amending the test case even further causes Pymoca to just outright fail in flattening:

```Modelica
package P
  ... unchanged ...
    class Level3 // Overrided max attribute
      // i.e. PostitivePin
      T.Voltage v(nominal=30, max=10.0);
    end Level3;
  end I;
  model M // Contains its own Level3 now
    // i.e. Resistor
    extends I.Level1(p.v(nominal=0), n.v(nominal=0));

    class Level3  // NEW, with yet another different max attribute
      T.Voltage v(nominal=30, max=30.0);
    end Level3;
  end M;

  class Level3
    T.Voltage v(nominal=30, max=20.0);
  end Level3;
end P;
```

Still the same `max=10.0` with `omc` though

kentrutan:
Whew! Fixed it without major tear-up. 

I was a bit concerned that more changes might be needed to the instantiation process. My first-pass look at comparing the procedure in Pymoca with the Modelica spec left me wondering. For instance, all that discussion on partial instantiation but Pymoca does things in-place, so do they match? (Spec v3.4 dated April 2017 brought in the better instantiation documentation so it was just a little before the initial code in the August 2017 multi-staged flattening commit 514402d.) Kudos to @jackvreeken for cracking a super-challenging problem with code that held up so far without a major rewrite.

I am going to hold off pushing the new code so I can do some more testing. I spent several days pulling my hair out on the original work with a lot of code thrown away, so it's worth at little more time on due diligence 😄.

jackvreeken:
Yeah, I spent a few weeks that summer figuring out the spec and rewriting Pymoca's codebase to properly flatten. The inheritance and applying of modifierse was so very wrong up until that point, except for the most basic of models. So your "several days" to figure it out means that it isn't just me that had a hard time with it ;) All that stuff is not really loaded in my "brain RAM" anymore though, but I'll help out where I can.

kentrutan:
Further testing has me playing the whack-a-bug game: fix for one case, breaks another. It appears that this is going to take some time and I may have to throw away yet another solution and start over.

jackvreeken:
Yeah, I was afraid of that... I'm not 100% sure that the existing test cases are correct, or at least, not sure that their assumed/tested result is equal to that of OMC. So if a case starts failing, do cross-check with OMC.

From what I understand, the symbol type resolution should _start_ with the new "original scope" you added, right? And not `extended_modified_class`. But I can very much imagine that other things start failing then... I asked Bing (GPT, whatever), and it says:

>According to the Modelica specification, a symbol’s type is resolved by looking up its name in the current scope and its enclosing scopes[1](https://specification.modelica.org/master/scoping-name-lookup-and-flattening.html). The current scope is determined by the flattening context, which consists of a modification environment and an ordered set of enclosing classes[1](https://specification.modelica.org/master/scoping-name-lookup-and-flattening.html). The name lookup follows these rules[1](https://specification.modelica.org/master/scoping-name-lookup-and-flattening.html):
>
> - If the name denotes a declared element of the current class or component, the type is that element.
>- If the name denotes an inherited element of the current class or component, the type is that element.
>- If the name denotes an element of a base class of the current class or component, the type is that element.
>- If the name denotes an imported element of the current class or component, the type is that element.
>- If none of the above applies, the name lookup is repeated in the next enclosing scope, until the global scope is reached.

kentrutan:
Amazing insight! You outlined exactly what happened. 🥇 I had actually started with lookup with the original symbol scope, but I changed it to the code in this PR when other test cases failed. Then it passed all tests at the time. Your code really is only as good as your test cases so thanks for helping to come up with a good one!

I'm somewhat impressed that the Bing/GPT search points to the right chapter when asked about "symbol" that is not part of the spec. It should have stopped with that because the details it gives are wrong. Maybe I should ask it to make `tree.py` comply 😆. I actually might try that just for fun. 🤔 

I'm going to try one more patch solution and if that doesn't work, I think it is going to have to be the more significant tear-up I alluded to earlier.

Here's a question for your "brain RAM" (or mag tape): Did you remember if you considered adding the "unnamed nodes" representing extends clauses into the instance tree during instantiation and what your thoughts were in why `build_instance_tree` can get away without doing that?

kentrutan:
Well I guess I'm not part of the elite that can use Bing/GPT. But I can get on a wait list 😆

jackvreeken:
> Did you remember if you considered adding the "unnamed nodes" representing extends clauses into the instance tree during instantiation and what your thoughts were in why build_instance_tree can get away without doing that?

Oof, I _think_ I just didn't understand what they were ("unnamed nodes" doesn't ring a bell even now), and that I just did the minimal viable thing to get models that were not working to work. As you say, my solution was only as good as my tests, and with the lack of parsing support for the Modelica standard library (that was not a high priority), I had to come up with all test cases myself.

If you do have a solution that implements these "unnamed nodes", I'll try to see if I can understand it, check it against the spec, and also try to poke holes in it ;)

kentrutan:
I am going to take a different direction for this fix that is closer to the language spec v3.5 section 5.6 and the discussion above. Other than the test case that evolved in this PR, the new implementation will have little in common with this one so I'm closing this PR.

PR #269, opened 2023-05-17
SymPy: Fix generator duplicate variable creation

kentrutan:
Removes duplicate SymPy symbols that the sympy generator backend creates for Modelica output variables. They are already created in the variables section of the Jinja template.

SymPy 1.12 aparently added a check for duplicate symbols. SymPy throws a `ValueError: duplicate symbols given` exception in `sympy.solve()` that causes `test_estimator` in `gen_sympy_test` to fail.

### DIFF
--- a/src/pymoca/backends/sympy/generator.py
+++ b/src/pymoca/backends/sympy/generator.py
@@ -138,9 +138,6 @@ class {{tree.name}}(OdeModel):
             {% endfor -%}}
 
         # outputs
-        {% if outputs_str|length > 0 -%}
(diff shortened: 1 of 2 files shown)


kentrutan:
I was able to duplicate the problem on my Mac and the fix passed (or so I thought?) but the CI failed on Ubuntu. I'll look into it later.

kentrutan:
Problem solved. I had selected the wrong python environment causing my local tests to pass when they should have failed. This new commit adds a corresponding change in the sympy runtime to go along with the previous change I had made to the Jinja template.

PR #275, opened 2023-08-01
Bump Python >= 3.8 and test to 3.11

kentrutan:
Python 3.7 is EOL.

Also, the recent CasADi 3.6.3 release supports Python up to 3.11. Since we do not pin the max Python and CasADi versions, we need to test up to Python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
           submodules: true
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: "3.11"
       - run: pip install tox
       - run: tox -e flake8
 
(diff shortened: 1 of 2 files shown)


jackvreeken:
LGTM. I also learned of "Renovate" recently (see #276) that can keep an eye on dependencies we have and create MRs with the latest versions of those dependencies. Then we can see if those pipelines fail, and we lag behind somewhere, or e.g. a new CasADi version introduces a bug.

Does that sound interesting to you as well, or do you rather want to manually keep track of bugs (i.e. run into them yourself when you update the dependencies, or when someone else does a fresh install and runs into issues).

kentrutan:
We don't test no minimum releases of dependencies or combinations in between with CI. It doesn't look like Renovate helps with that either, so I don't see the value.

We do need to update ANTLR. I have been using 4.9 for a while now and was relatively easy to update. It has some speedups and works better with VSCode. [Speed Antlr Tool](https://github.com/amykyta3/speedy-antlr-tool) only supports up to 4.11, so I'll take a look at how difficult it is to go that far.

PR #277, opened 2023-08-02
CI/CD: Add GitHub release action

kentrutan:
This adds an action to create a GitHub release after the PyPI release actions so that the repo will reflect the release on PyPI. 

It has not been tested since I don't know how to do that without going live.

EDIT
I should probably add that there is an [Official PyPA release action](https://github.com/marketplace/actions/pypi-publish) that might be good to look at for future updates to the CD. See, for instance, the [flask publish workflow](https://github.com/pallets/flask/blob/main/.github/workflows/publish.yaml) which also creates a GitHub release using a more manual method than the action I used in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,3 +85,8 @@ jobs:
       run: |
         python setup.py sdist bdist_wheel
         twine upload dist/*
+    - uses: softprops/action-gh-release@v1
+      with:
+        files: dist/*
+        generate_release_notes: true
+        


jackvreeken:
I think to test this, I'd recommend you make a branch "add-github-release" in this repo. 
Then you can just make a tag on that branch of the type 0.10.0.dev0 (increment last digit every time you test a new commit + tag). You definitely don't have to merge to master just yet, you can test this properly before merging.

On the note of making releases, over in https://github.com/Deltares/libzsf/commit/02913b9f86393c74bbd2615298603eb749ca6f02 I added something similar (that commit moved CI/CD from GitLab to GitHub), and I used `ncipollo/release-action@v1`. Adapting a little for our use case here (haven't tested):

```yml
...
        TWINE_USERNAME: __token__
        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
      run: |
        python setup.py sdist bdist_wheel
        twine upload dist/*

# START OF NEW STUFF
    - uses: actions/upload-artifact@v3
      with:
        name: dist
        path: dist
        retention-days: 5

  deploy-github-release:
    needs:
      - deploy
    runs-on: ubuntu-latest
    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags') }}
    permissions:
      contents: write
    steps:
    - uses: actions/download-artifact@v3
      with:
        path: dist/
    - uses: ncipollo/release-action@v1
      with:
        artifacts: "dist/*.whl,dist/*.tar.gz"
        generateReleaseNotes: true
        draft: true
```

kentrutan:
Thanks for the advice. Will do!

PR #271, opened 2023-07-27
CasADi: Fix caching of delay expression with constant delay

jackvreeken:
Closes #270

PR #279, opened 2023-08-06
0.9.0 -> 0.9.1: Bug fixes to include

jackvreeken:
- [x] #271


jackvreeken:
With the amount of changes required to e.g. the CI/CD, running on unsupported Python versions, version pinning of dependencies, I think I'm just gonna close this MR and suggest people use 0.10.* instead (and pin the exact version they want).

PR #280, opened 2023-08-12
playing around with casadi cse

jackvreeken:


jackvreeken:
WIll close the PR. Don't really have a good test case or set of benchmarks to look at how this might impact performance. I'm _hoping_ it does of course, but we need hard numbers before merging something like this.

PR #281, opened 2023-08-16
Style: Format code with black

jackvreeken:


jackvreeken:
@kentrutan It's been bugging me that we don't use black for Pymoca, leading to inconsistent style in the code. Not a big deal, just one of my pet peeves I guess. This is a monster commit, but it really is just the black command and nothing else.

Doing it all in one commit makes it easy to still use e.g. git blame to find out who/what/why changed a line. I'd also be in favor to also add this monster-style commit hash to an ignore file so `blame` still works nicely, see https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view .

kentrutan:
I think this is a good idea in general and it bugs me too, so I'm glad you are working on it! 👏 

I am not sure about all the quote changes. I agree that being consistent with quotes is good, but double quotes everywhere? There is an option to leave quotes alone, but that would mean we need to manually make quotes consistent. 😕 Let me peruse the code and think about it a bit more when I have more time.

I agree with ignoring this commit for git blame.

Some things I noticed so far you might want to consider:

1. Should _version.h be ignored since it is generated by versioneer.py?
2. Should you specify all versions of Python supported per the black docs recommendation?
3. You probably noticed that the CI flagged max line length is 99, not 100.

jackvreeken:
1. Hmm, maybe. But it's only generated _once_, and we're never going to touch it again. I don't remember what I did in the past with this file elsewhere. 
2. Good catch! I guess it's also about time to work on a pyproject.toml file .
3. I'm not sure what setup.cfg I was looking at, but it wasn't that of Pymoca's. I should probably deal with all those exclusions there in this PR as well.

kentrutan:
That would be awesome if you get us more up to date using pyproject.toml!

On my fork there is a commit from March were I changed to use the latest pip-installed versioneer. I don't remember why I updated, but I had my reasons :smile:. It appears there are some [more recent versioneer updates](https://github.com/python-versioneer/python-versioneer/releases) related to pyproject.toml and Python 3.11. How about I update to the latest and make a PR?

kentrutan:
Over on #282 there is an updated versioneer and a starter pyproject.toml with only what is needed for versioneer.

jackvreeken:
Turns out both flake8 and tox don't really deal with pyproject.toml yet. But that shouldn't stop us from moving into that direction regardless imo.

src/pymoca/ast.py
```diff
@@ -220,69 +226,88 @@ def concatenate(self, arg: 'ComponentRef') -> 'ComponentRef':
 class Expression(Node):
     def __init__(self, **kwargs):
         self.operator = None  # type: Union[str, ComponentRef]
-        self.operands = []  # type: List[Union[Expression, Primary, ComponentRef, Array, IfExpression]]
+        self.operands = (
+            []
+        )  # type: List[Union[Expression, Primary, ComponentRef, Array, IfExpression]]
```
kentrutan, line 231: Why did it break this line (was only 87 long)?

src/pymoca/ast.py
```diff
@@ -293,99 +318,115 @@ def __init__(self, **kwargs):
...
+        "value",
+        "min",
+        "max",
+        "start",
+        "fixed",
+        "nominal",
+        "unit",
+        "quantity",
+        "displayUnit",
+    ]
```
kentrutan, line 415: While vertically challenged, this does make it a little easier to peruse with diff tools that don't do character-by-character.

src/pymoca/ast.py
kentrutan: Perhaps it is so that you can easily see in a diff changes to only type hints change vs changes in the functioning code?

src/pymoca/ast.py
```diff
@@ -631,8 +692,13 @@ def _find_class(self, component_ref: ComponentRef, search_parent=True, search_im
                 else:
                     raise ClassNotFoundError("Could not find class '{}'".format(component_ref))
 
-    def find_class(self, component_ref: ComponentRef, copy=True,
-                   check_builtin_classes=False, search_imports=True) -> 'Class':
+    def find_class(
```
kentrutan, line 695: I guess the black format highlights ugly functions that maybe should be cleaned up haha!

src/pymoca/ast.py
```diff
@@ -839,4 +907,4 @@ def update_parent_refs(self) -> None:
         self._update_parent_refs(self)
 
     def __repr__(self):
-        return '{}(classes={!r})'.format(type(self).__name__, self.classes)
+        return "{}(classes={!r})".format(type(self).__name__, self.classes)
```
kentrutan, line 910: I was concerned black migh mess up `__repr__`'s in the ast that should use single quotes, but it appears all is well. In hindsight, I think even if I had not used `!r` everywhere, black would respect single quotes inside double quotes.

src/pymoca/parser.py
```diff
@@ -346,60 +347,42 @@ def exitExpression_if(self, ctx: ModelicaParser.Expression_ifContext):
...
-            operands=[self.ast[e] for e in ctx.expr()]
-        )
+        self.ast[ctx] = ast.Expression(operator="or", operands=[self.ast[e] for e in ctx.expr()])
 
     def exitExpr_signed(self, ctx: ModelicaParser.Expr_signedContext):
-        self.ast[ctx] = ast.Expression(
-            operator=ctx.op.text,
-            operands=[self.ast[ctx.expr()]]
-        )
+        self.ast[ctx] = ast.Expression(operator=ctx.op.text, operands=[self.ast[ctx.expr()]])
```
kentrutan, line 385: Well black prefers less readable with these `Expr`*, haha. I guess this shows a human can sometimes make better decisions than hard-wired formatting rules.

kentrutan:
I'm OK with using black. While I'm not thrilled with some of its choices, I think the benefits outweigh some of the ugliness. Please proceed and thanks for taking the initiative to move us along! :rocket:

src/pymoca/ast.py
jackvreeken: I guess, we should probably make type annotations proper Python 3 style as well, not as comments. There's a whole bunch of legacy stuff in the code base still

src/pymoca/parser.py
jackvreeken: I think we can force what you want by adding a comma after the `operands=[self.ast[ctx.expr()]]` line

test/generated/Aircraft.py
kentrutan, line 1: Why are files in test/generated included in the PR diffs and look like they are formatted with black? They should be skipped and not part of this PR since running tests will generate the format encoded in the jinja template.

kentrutan: It looks like they should have been skipped according to pyproject.toml, so maybe unexpectedly snuck in there somehow.

src/pymoca/ast.py
kentrutan: Marking this as resolved. Did you record your thought somewhere (maybe an Issue?) about changing to Python 3 type annotations?

kentrutan (changes requested):
This is going to be very nice, especially when there are more collaborators. I'm good with you going ahead with the merge after addressing new comments. Thanks!

src/pymoca/tree.py
```diff
@@ -161,8 +211,12 @@ def skip_child(self, tree: ast.Node, child_name: str) -> bool:
...
         """
-        if isinstance(tree, ast.Class) and child_name == 'parent' or \
-                isinstance(tree, ast.ClassModificationArgument) and child_name in ('scope', '__deepcopy__'):
+        if (
+            isinstance(tree, ast.Class)
+            and child_name == "parent"
+            or isinstance(tree, ast.ClassModificationArgument)
+            and child_name in ("scope", "__deepcopy__")
+        ):
             return True
```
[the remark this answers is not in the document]
jackvreeken: Black is not magic in that sense, but I'll add parentheses for clarity. Indeed it's otherwise a bit lost that `and` precedes `or`.

test/generated/Aircraft.py
jackvreeken: Good catch! Fixed :)

PR #282, opened 2023-08-17
Update versioneer

kentrutan:
Remove the versioneer copy and just use pip installed version. Update embedded code generated by `versioneer install --no-vendor`.

Versioneer, Python 3.11, and/or latest setuptools now apparently require the use of pyproject.toml. We are planning on moving to that anyway so this is a start.

This has only been tested with my Python 3.11 environment and just doing a `python -m build`, so more testing is needed.


jackvreeken:
Ah, I was a bit confused at first, but indeed this commit changes versioneer from "vendor" mode to "non-vendor" mode, i.e. not including a copy of it. I don't think that was an option way back when, so good find!

I assume you're going to get rid of the stuff in `setup.cfg`? That's about the only thing I can see that still makes this "draft", otherwise LGTM

kentrutan:
> I assume you're going to get rid of the stuff in `setup.cfg`? That's about the only thing I can see that still makes this "draft", otherwise LGTM

I removed the extraneous comments in setup.cfg. Is that the "stuff" you mentioned? All the other setup.cfg migration I hoped you would do per the discussion over in #281.

kentrutan:
While looking into all the discussion over the years on tools supporting pyproject.toml (or not), I stumbled on [setuptools_scm](https://pypi.org/project/setuptools-scm/) which might be a simpler alternative to Versioneer in a pyproject.toml world. Does it make sense to migrate to that along with the pyproject.toml migration? Any reason why Versioneer is better for Pymoca?

kentrutan:
I'll leave this to you @jackvreeken to merge if you wish or close if you want to pursue setuptools_scm instead. I have tested it round trip on testpypi. CI did hiccup on a mac build (timeout?) but I restarted and it passed as you can see.

jackvreeken:
Did a quick check what the results are on a clean and dirty work tree, and it still all works just fine.

As far as setuptools_scm vs versioneer is concerned, both are entirely fine by me. It seems the jury is still out on which to choose, especially now that versioneer development has picked up again. I'll leave it to you then to either rebase this and merge, or see what setuptools_scm would look like.

kentrutan:
Honestly, I need to get back to instantiation rewrite, so if I'm doing the work, Versioneer it is for now! 

I agree that Versioneer looks better now that it is being updated, but I'm guessing setuptools_scm is probably the better choice long run due to its connection to PYPA.

jackvreeken:
> @jackvreeken what is your setup with Black to prevent the above CI fail from happening? I added black on save in Visual Studio Code, but that doesn't help with this one where Versioneer updated `src/pymoca/__init__.py` with a format black doesn't like. Do you use a pre-commit hook somewhere to run black?

Hmm, I just run `black src test tools setup.py` before commiting, but I do both those just in PowerShell (on Windows).

PR #285, opened 2023-08-24
CasADi/Simplify: Fix `resolve_parameter_values` for list/array values

jackvreeken:
The symbol "value" was not instantiated when the parameter value was a list.

Furthermore, we were only looking at lists and not also at NumPy arrays. We force convert from list to array in this if-branch, and also in some places in the code. We therefore have to make sure the condition catches both lists and np.ndarrays.

### DIFF
--- a/test/gen_casadi_test.py
+++ b/test/gen_casadi_test.py
@@ -3015,6 +3015,56 @@ class GenCasadiTest(unittest.TestCase):
         casadi_model.simplify(compiler_options)
         self.assertEqual(ineg.value, -1.0)
 
+    def test_resolve_parameter_values_lists(self):
+        """Test that `resolve_parameter_values` works for lists/arrays"""
+        txt = """
+            model A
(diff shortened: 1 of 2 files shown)


jackvreeken:
Was initially also working on extending the test with `replace_parameter_expressions`, but ran into some other bugs there.

test/gen_casadi_test.py
```diff
@@ -3015,6 +3015,56 @@ def test_signed_expression(self):
         casadi_model.simplify(compiler_options)
         self.assertEqual(ineg.value, -1.0)
 
+    def test_resolve_param_expressions(self):
+        """Test that both + and - prefix operators work in expressions"""
```
kentrutan, line 3019: Looks like the comment needs an update after copy/paste. (Not marked draft, so assuming you were ready for review.)

kentrutan: Good catch on the bug and nice to see more test coverage! 

Looks good to me other than the above.

PR #289, opened 2023-08-25
Fixes for black formatting

jackvreeken:


jackvreeken:
Whoopsie, forgot to actually add black to CI/CD pipeline 🤦‍♂️

.github/workflows/ci.yml
```diff
@@ -26,8 +26,21 @@ jobs:
       - run: pip install tox
       - run: tox -e flake8
 
+  style-black:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
```
kentrutan, line 35: Does black really need any of the submodules? I also wonder if flake8 needs them (perhaps for xml?).

kentrutan (changes requested):
Up to your judgement on addressing the question and please go ahead and merge when you're done.

.github/workflows/ci.yml
jackvreeken: No, they don't need it afaik, they don't even need an install.

Only reason it's there is because it was there for flake8.

I'll add a separate commit that deletes it for flake8, and then add a black stage without

PR #292, opened 2023-08-29
Remove generated test files from repository

### LINKED ISSUE #291: Clean up test/generated/*
I'm not entirely sure why these files are part of version control. They are generated each test run, and current generation differs from the files generated in the past (so git thinks they're modified). Can we remove them, and add them to .gitignore?

jackvreeken:
These files are generated when running the tests. Typically, their style does not adhere to black's standards, and thus this leads to seeing them in the modified section when running `git status`, and potentially accidentally committing them. As there is no need for them to be version controlled, we remove them.

Note that we need to keep the folder around for the tests to succeedi (they do not make it if it does not exist), so we use a .gitkeep file in there.

Closes #291


kentrutan (approved):
Totally agree with this one! It has been a pain in the neck for me too, so I had removed it on my fork. Thanks for making it official!

PR #294, opened 2023-09-18
Parser: Update ANTLR to 4.13.1

kentrutan:
Our ANTLR 4.7.* runtime uses `typing.io` that will be deprecated in the imminent Python 3.12 release, so we are updating to the latest ANTLR available at this time. There is no measurable speed improvement with this update in Python 3.11.

We plan to integrate https://github.com/amykyta3/speedy-antlr-tool at some point. It currently only mentions ANTLR support up to 4.11 so hopefully it is forward compatible or it is updated by the time we get to that project. If not, I can submit a PR as a last resort.


jackvreeken:
LGTM, feel free to go ahead and merge

PR #295, opened 2023-09-18
Compiler: Delay backend imports

kentrutan:
This allows the compiler to be used when sympy and/or casadi are not installed. A use case is testing just the parse and/or flatten steps. It should also speed up the startup of the compiler a little, even when one of the backends is used, and becomes more and more important as additional backends are added.

The implementation simply moves the sympy and casadi backend imports to where they are used.


jackvreeken:
LGTM! Feel free to go ahead and merge

PR #296, opened 2023-09-19
Style: Use pre-commit

kentrutan:
Wanted something to automatically check for style conformance before pushing to GitHub and the [pre-commit Python package](https://pypi.org/project/pre-commit/) seems to do the trick.

In addition, since there are GitHub actions for pre-commit, I propose changing the CI to use pre-commit also so local and GitHub configuration for style checks are consistent and in once place. This is the 2nd commit in this PR, so it can be dropped if we decide not to do that.

If we adopt for GitHub CI, there is a [pre-commit GitHub action](https://github.com/marketplace/actions/pre-commit) and there is also a separate [pre-commit CI service](https://pre-commit.ci) that says it is free for open source.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+repos:
+  # Using this mirror lets us use mypyc-compiled black, which is about 2x faster
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 23.9.1
+    hooks:
+      - id: black
+        args: [--check, --diff]
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        # TODO: Commenting out max supported Python version until we pin it for Pymoca
+        # language_version: python3.11
+  - repo: https://github.com/PyCQA/flake8
+    rev: 6.1.0
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          [flake8-bugbear, flake8-comprehensions, flake8-import-order]
+        # This exclude list should be same as for black in pyproject.toml
+        # See https://pre-commit.com/#regular-expressions for this syntax
+        exclude: |
+          (?x)^(
+            src/pymoca/generated/.*|
+            src/pymoca/_version.py|
+            test/generated/.*|
+            test/libraries/.*
+          )$
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ on:
     - cron: "6 6 * * 3"
 
 jobs:
-  style-flake8:
+  pre-commit:
     runs-on: ubuntu-latest
 
     steps:
@@ -21,22 +21,10 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
-      - run: pip install tox
-      - run: tox -e flake8
-
-  style-black:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
-        with:
-          python-version: "3.11"
-      - run: pip install tox
-      - run: tox -e black
+      - uses: pre-commit/action@v3.0.0
 
   test:
-    needs: [style-flake8, style-black]
+    needs: pre-commit
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -98,4 +86,4 @@ jobs:
       with:
         files: dist/*
         generate_release_notes: true
-        
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ parentdir_prefix = "pymoca-"
 [tool.black]
 line-length = 100
 target-version = ['py38', 'py39', 'py310', 'py311']
-exclude = '''
+force-exclude = '''
 	(
 		  src/pymoca/generated
 		| src/pymoca/_version.py
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    flake8,py,coverage
+    py,coverage
 
 
 [testenv]
@@ -17,22 +17,3 @@ deps =
 passenv = GITHUB_ACTIONS
 commands =
     pytest --cov --cov-report=xml test {posargs}
-
-
-[testenv:flake8]
-skip_install = true
-application-import-names = pymoca,test
-deps =
-  flake8
-  flake8-bugbear
-  flake8-comprehensions
-  flake8-import-order
-commands = flake8 src test tools setup.py
-
-
-[testenv:black]
-skip_install = True
-deps =
-    black >= 23.7.0
-commands =
-    black --check --diff src test tools setup.py


.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 23.9.1
+    hooks:
+      - id: black
+        args: [--check, --diff]
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        # language_version: python3.11
```
jackvreeken, line 12: Why is this commented out?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+repos:
+  # Using this mirror lets us use mypyc-compiled black, which is about 2x faster
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 23.9.1
```
jackvreeken, line 4: Does this mean we're now pinning versions?

jackvreeken (changes requested):
I don't know what you mean with:

> I don't see anything in the pre-commit docs that shows how fixes could
be automated locally, but there is a GitHub action pre-commit.ci that
fixes things automatically.

Isn't the whole point of pre-commit hooks that fixes _are_ automated locally?

jackvreeken:
I was also recently made aware of Ruff as an alternative for flake8. It's very fast, and has support for pyproject.toml

kentrutan:
> I don't know what you mean with:
> 
> > I don't see anything in the pre-commit docs that shows how fixes could
> > be automated locally, but there is a GitHub action pre-commit.ci that
> > fixes things automatically.
> 
> Isn't the whole point of pre-commit hooks that fixes _are_ automated locally?

It is possible to have pre-commit go ahead and apply black by taking out the `--check` argument in .pre-commit-config.yaml, but that doesn't seem to work with tox. Also, it just creates an unstaged version of the file that you still have to manually stage to proceed with the commit.

For flake8, is there a way to have it automatically apply fixes? pre-commit is really just a tool to help with creating and updating git pre-commit hooks that work with tools like black and flake8. It's up to you or the called tool to make the changes.

.pre-commit-config.yaml
kentrutan: We don't have the max Python version pinned in setup.py, so why pin it here? Just another thing to have to keep updated. I expect when we get to Pymoca v1.0, we will want to pin a lot more things, including max Python version.

.pre-commit-config.yaml
kentrutan: Good question. I did not try to see if we can omit the rev. @jackvreeken If you agree that we should proceed with pre-commit, I'll give that a try and report back. 

Also, what do you think about using the GitHub action? I don't think I can implement that with my current level of access in the Pymcoa organization. If we don't use it on GitHub, then we can partially automate the local black fix ala https://github.com/pymoca/pymoca/pull/296#issuecomment-1729753202. Without access to changing GitHub CI action myself, it is impossible to test, so if you don't have time to work on it, then I can just drop the CI part of this PR. My main motivation was to prevent style-related CI fails on PRs and the local use of pre-commit does it.

.pre-commit-config.yaml
jackvreeken: Well, then the question is, why are you leaving the comment above the language-specification line

kentrutan: > Well, then the question is, why are you leaving the comment above the language-specification line

Because I didn't want to remove the other comments that pre-commit generated and for future reference when we do pin versions. You make a good point, I should put in a comment to explain this.

.pre-commit-config.yaml
jackvreeken: >  I don't think I can implement that with my current level of access in the Pymoca organization. 

I thought I made you maintainer, I'll see if there's a higher level

.pre-commit-config.yaml
jackvreeken: Makes sense, I thought it was copy-paste from some example, but if it's generated indeed just keep it

kentrutan:
@jackvreeken I keep running into pushing commits with black format errors, so would like to at least get the `.pre-commit-config.yaml` commit into master. Since you gave me more permissions, I can take a swing at changing the CI also if you agree.

.pre-commit-config.yaml
kentrutan: > I did not try to see if we can omit the rev.

I confirmed that the rev key is required. pre-commit.ci has an option that will automatically update the revs to latest and commit the update on a weekly basis. It also has an option to automatically fix files that aren't black conformant, but I'm not thrilled with the idea of polluting the history with bot commits. If we scale up to more collaborators, it might make sense to revisit this.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+        # language_version: python3.11
+  - repo: https://github.com/PyCQA/flake8
+    rev: 6.1.0
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          [flake8-bugbear, flake8-comprehensions, flake8-import-order]
+        # This exclude list should be same as for black in pyproject.toml
+        # See https://pre-commit.com/#regular-expressions for this syntax
+        exclude: |
```
jackvreeken, line 22: I don't quite get this. Isn't this covered by the file list in `setup.cfg`?

jackvreeken (changes requested):
Double declaration of file list is a bit weird.

Also, really tempted to see if we can move to Ruff instead of flake8, and how that works. I have no experience with it, but I like speed.

.pre-commit-config.yaml
kentrutan: > Double declaration of file list is a bit weird.

I totally agree. It is caused by flake8 not having the force-exclude option like black. Another option is to only specify in the (admittedly ugly) pre-commit exclude lists and then only run the style tools through pre-commit. That might cause problems for use of the tools through IDEs.

> 
> Also, really tempted to see if we can move to Ruff instead of flake8, and how that works. I have no experience with it, but I like speed.'

Ruff does have force-exclude, so would take care of the double file list issue. How tempted are you to implement? I personally would like to get back to working on the Modelica issues rather than style-related stuff. Do you want to take up the flake8 replacement with Ruff perhaps as a follow-on to this while simultaneously putting all of the exclude file lists into pyproject.toml. Adding a change to Ruff in this pre-commit PR is a bit of a tangent.

Speaking of speed, it looks like there is a lot of room for speeding up the CI by studying CI run logs. One I notice is the install of jupyterlab when there are no tests of notebooks. Also, can't we use caching? The pre-commit action seems to automatically use caching.

jackvreeken: In that case, just leave the double exclude stuff in there. I tend to like working on style stuff more than you, so I'll pick up the Ruff proof-of-concept some day soon^TM (no promises).

> Speaking of speed, it looks like there is a lot of room for speeding up the CI by studying CI run logs. One I notice is the install of jupyterlab when there are no tests of notebooks. 

That's because we have `extras = all`, instead of `casadi,sympy,lxml`. Didn't notice it being a cause of slowness, but would be a quick fix. Also outside the scope of this MR of course.

Anyway, with my one "complaint" addressed, LGTM now.

kentrutan: > I'll pick up the Ruff proof-of-concept some day soon^TM (no promises).

Sorry you trademarked "some day soon". I could use that one. 😆

kentrutan: BTW @jackvreeken, I really do like working on tooling. Just want to show more progress on Modelica instantiation/flattening right now while you are working on it also. You: 🛩️ Me: 🐌

PR #303, opened 2023-09-23
Use setuptools_scm instead of versioneer

jackvreeken:
We want to be able to override the version number, which is not supported by versioneer.

We need this for e.g. complete coverage testing of #287 

Custom local tag form with `dirty` at the end, instead of a `dYYYYMMDD`. The latter makes it hard to detect from within the code if it's dirty or not.

WIP because following two don't match yet for some reason

```
python -c "import pymoca; print(pymoca.__version__)"
```
outputs "0.10.1.dev20+unknown"

```
python -m setuptools_scm
```
outputs "0.10.1.dev20+g3123894"

- [ ] Test that version override actually works
- [x] Why does requested version number not change when e.g. tagging. So I would expect `python -c "import pymoca; print(pymoca.__version__)"` -> `git tag 0.10.99` -> `python -c "import pymoca; print(pymoca.__version__)"` to give different strings as a result (as it does with versioneer). This does not seem to be the case. Why?


jackvreeken:
I think I'll just close this, as dynamic / local version strings are still too difficult if not outright impossible with setuptools_scm.

jackvreeken:
related: https://github.com/maresb/hatch-vcs-footgun-example

jackvreeken:
I'll reopen this for now as a reminder, but my plan is to 👍 
- Try hatchling (instead of setuptools)
- Use hatch-vcs instead of versioneer (uses setuptools_scm under the hood)
- Use the footgun example, because we rely on "dirty" detection for local dev environments.

Hopefully that'll work 🤞

(tagging just for the FYI @kentrutan , you don't have to do anything)

PR #305, opened 2023-09-26
Tree: Fix inherited type value modification

### LINKED ISSUE #300: Error in 0.10.0 when setting alias type
Pymoca 0.10.0 transfer_model seems to crash when setting an alias type.

Minimal example: a package file
```Modelica
package Package
  type Length = Real;
  type Height = Length;
end Package;
```
with submodel
```Modelica
model SubModel
  input Package.Height H;
end SubModel;
```
and model
```Modelica
model Model
  SubModel var(H=5);
equation
  // var.H = 5;
end Model;
```
and python file
```python
from pathlib import Path
import pymoca.backends.casadi.api

model_folder = Path(__file__).resolve().parent

model = pymoca.backends.casadi.api.transfer_model(model_folder, "Model")
```
all in the same folder. When running the python file with python 3.9, pymoca 0.10.0, casadi 3.6.3, it gives the following error message:

> pymoca.tree.ModificationTargetNotFound: Trying to modify symbol value, which does not exist in class Package.Height

It works fine when replacing `Height` with `Length` in the submodel file. It also works fine when replacing `var(H=5)` with `var` and uncommenting the equation in the model file.

kentrutan:
This fixes a problem where modifying the value of a symbol of a type that is inherited from another type would fail during instantiation with a `pymoca.tree.ModificationTargetNotFound`.

This also expands the "Processing failed for symbol" message given when symbol type instantiation fails in `build_instance_tree`. It now includes the message from the underlying exception to give a more specific reason for the failure.

This is related to 4dd9048 that fixed inherited types for two test cases using SI units in MSL 3 and 4, but they did not include testing value *modification* of the inherited types.

Closes #300.

### DIFF
--- a/src/pymoca/tree.py
+++ b/src/pymoca/tree.py
@@ -427,9 +427,12 @@ def build_instance_tree(
 
     # Check that all symbol modifications to be applied on this class exist
     for arg in extended_orig_class.modification_environment.arguments:
-        if arg.value.component.name not in extended_orig_class.symbols:
+        if (
+            arg.value.component.name not in extended_orig_class.symbols
+            and arg.value.component.name not in ast.Symbol.ATTRIBUTES
+        ):
             raise ModificationTargetNotFound(
-                "Trying to modify symbol {}, which does not exist in class {}".format(
+                'Trying to modify symbol "{}", which does not exist in class {}'.format(
                     arg.value.component.name, extended_orig_class.full_reference()
                 )
             )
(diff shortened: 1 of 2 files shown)


jackvreeken:
Has definitely been a while since I looked at the flattening/lookup code, but I think I understand the changes (I stepped through them with the debugger). So LGTM, feel free to merge whenever :)

One question I have for you related to the future fully spec-compliant instantiation, are you getting rid of the special variable name "__value"? It's probably an ugly workaround I came up with to make things work, but if someone out there actually defines their own variable with that name I don't know what will happen 😅

kentrutan:
Thanks for checking it out.

I expect that I'll be able to get rid of the special `__value` *and* `__builtin` code, but I'm working on a first step of making the instantiation and flattening steps line up better with the spec by dicing up the current code with a minimum of changes. Next steps are cleaning it up. I'm in the middle of fleshing it out, so the plan is fluid. I'll try to get to a point where something is working (or at least half-baked and understandable :-) and put out a draft PR for some feedback. If I can only find time to work on it 😉.

Issue #266, opened 2023-03-29
Flattening MSL 4 `Resistor` fails


kentrutan:
Flattening `Modelica.Electrical.Analog.Basic.Resistor` from Modelica Standard Library v4.0.0 fails. The root cause is the symbol type name `PositivePin` is referenced in an inheritance hierarchy two levels deep but the original scope is lost in `tree.flatten_extends()`.

Here is a test case similar to MSL 4 Resistor that fails with `IndexError: Processing failed for symbol "P.M.p"`. It has some additional twists added to check for correct lookup scope for `Level3` (`max` should be `10.0`) and correct modification (`nominal` should be `0.0`).
```modelica
package P
  package T
    class Voltage = Real(nominal=1);
  end T;
  package I
    class Level1
      // i.e. OnePort
      extends Level2(p.v(nominal=10), n.v(nominal=10));
    end Level1;
    class Level2
      // i.e. TwoPin
      Level3 p(v(nominal=20)), n(v(nominal=20));
    end Level2;
    class Level3
      // i.e. PostitivePin
      T.Voltage v(nominal=30, max=10.0);
    end Level3;
  end I;
  model M
    // i.e. Resistor
    extends I.Level1(p.v(nominal=0), n.v(nominal=0));
    class Level3 // Test for wrong lookup scope with max=30.0
      T.Voltage v(nominal=30, max=30.0);
    end Level3;
  end M;
  class Level3 // Test for wrong lookup scope with max=20.0
    T.Voltage v(nominal=30, max=20.0);
  end Level3;
end P;
```
See #265 for additional discussion.

kentrutan:
Closed by #307

PR #306, opened 2023-09-26
Tree: Fix storing constant references in non-elementary typed equations

### LINKED ISSUE #301: Error for 0.10 when setting typed var with package constant
Pymoca 0.10.0 transfer_model seems to crash when setting a typed variable with a constant from a package.

Minimal example: a package file
```Modelica
package Package
  type Length = Real;
  constant Real KM_TO_M = 1000.0;
end Package;
```
with model
```Modelica
model Model
  constant Real KM_TO_M = 1000;
  Package.Length distance_km = 1.0;
  Package.Length distance_m = Package.KM_TO_M * distance_km;
  // Package.Length distance_m = KM_TO_M * distance_km;
end Model;
```
and python file
```python
from pathlib import Path
import pymoca.backends.casadi.api

model_folder = Path(__file__).resolve().parent

model = pymoca.backends.casadi.api.transfer_model(model_folder, "Model")
```
all in the same folder. When running the python file with python 3.9, pymoca 0.10.0, casadi 3.6.3, it gives the following error message:

> KeyError: 'Package'

It works fine when `distance_m` is a `Real` or when using the constant `KM_TO_M` from the model file.

jackvreeken:
For the added test case (see the diff), what was happening is that we added the resolved constant to the wrong class. We added it to the Package.Length class, i.e. the InstanceClass of the type of `distance_m`. Basically what was happening in the example was

1. Entering Model
2. Entering Length
3. Finding reference to constant, changing reference and storing new constant symbol
4. Exiting Length - popping off the stored constant symbols
5. Exiting Model

The reason for this happening was that it was not envisioned at the time of the offending code that we would be entering InstanceClasses other than those part of the lookup tree hierarchy.

Instead, we want to add the stored constant symbols to `Model`. So what we currently implement for the last two steps above is:

4. Exiting Length - skipping popping
5. Exiting Model - popping off the stored constant symbols

Closes #301

Things to test:
- [ ] Extends a redeclared class. E.g. the AT = BT, but then somehow extending from AT. We do a find_class with copy now in `flatten_extends`, but maybe shouldn't then?

Things for future:
- [ ] Would be really awesome if we could just compare output / trees with `omc -i=...` (or reference flattened case)
- [ ] Would be nice if we can handle the `assert(Util.compareReal ...)` stuff like in ImportScopeType . Then we can just use the already existing .mo files in the modelica test library as test cases.

### DIFF
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -250,6 +250,20 @@ class AttributeRef(ComponentRef):
         super().__init__(name=component_ref.name)
 
 
+class ForLoopIndexRef(ComponentRef):
+    def __init__(self, component_ref: ComponentRef):
+        # TODO:
+        # Had to disable checks, because we also loop over annotations where
+        # other things than "value/nominal/..." are referenced
+        # assert len(component_ref.child) == 0
+        # if component_ref.name not in Symbol.ATTRIBUTES:
+        #     a = 1
+        # assert component_ref.name in Symbol.ATTRIBUTES
+        # assert component_ref.indices == [[None]]
+
+        super().__init__(name=component_ref.name)
+
+
 class SymbolTypeRef(ComponentRef):
     def __init__(self, component_ref: ComponentRef):
         assert len(component_ref.child) == 0
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -829,7 +829,7 @@ class Class(Node):
 
                 # Found a symbol. Continue lookup on type of this symbol.
                 if isinstance(s.type, InstanceClass):
-                    return s.type._find_symbol(ComponentRef.from_tuple(t[1:]), False)
+                    return s.type._find_symbol(ComponentRef.from_tuple(t[1:]), False, in_original_scope)
                 elif isinstance(s.type, ComponentRef):
                     node = self._find_class(s.type)  # Parent lookups is OK here.
                     return node._find_symbol(ComponentRef.from_tuple(t[1:]), False)
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -852,6 +852,20 @@ class Class(Node):
                     else:
                         raise NonConstantSymbolFoundError(f"Symbol '{component_ref}' found in enclosing scope, but is not a constant.")                   
             except KeyError:
+                # Try imports
+                if "*" in self.imports:
+                    for package_ref in self.imports["*"].components:
+                        imported_comp_ref_class = package_ref
+                        try:
+                            c = self._find_class(imported_comp_ref_class)
+                            sym = c.find_symbol(component_ref)
+                            if "constant" in sym.prefixes:
+                                return sym
+                            else:
+                                raise NonConstantSymbolFoundError(f"Symbol '{component_ref}' found in imports, but is not a constant.")
+                        except (KeyError, ClassNotFoundError):
+                            pass
+
                 if search_parent and self.parent is not None:
                     return self.parent._find_symbol(component_ref)
                 else:
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -248,7 +248,7 @@ class AttributeRef(ComponentRef):
 class SymbolTypeRef(ComponentRef):
     def __init__(self, component_ref: ComponentRef):
         assert len(component_ref.child) == 0
-        assert component_ref.name == "Real"
+        assert component_ref.name in Class.BUILTIN
         assert component_ref.indices == [[None]]
 
         super().__init__(name=component_ref.name)
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -238,9 +238,14 @@ class ComponentRef(Node):
 
 class AttributeRef(ComponentRef):
     def __init__(self, component_ref: ComponentRef):
-        assert len(component_ref.child) == 0
-        assert component_ref.name in Symbol.ATTRIBUTES
-        assert component_ref.indices == [[None]]
+        # TODO:
+        # Had to disable checks, because we also loop over annotations where
+        # other things than "value/nominal/..." are referenced
+        # assert len(component_ref.child) == 0
+        # if component_ref.name not in Symbol.ATTRIBUTES:
+        #     a = 1
+        # assert component_ref.name in Symbol.ATTRIBUTES
+        # assert component_ref.indices == [[None]]
 
         super().__init__(name=component_ref.name)
 
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -1041,6 +1041,8 @@ class Tree(Class):
     The root class.
     """
 
+    # TODO: Where do we check if the number and type of arguments is correct?
+    # What does OMC do on this front?
     BUILTIN_FUNCTIONS = {
         # See Modelica Specification:
         # Section 3.7 Built-in Intrinsic Operators with Function Syntax
@@ -1097,6 +1099,39 @@ class Tree(Class):
         "edge",
         "change",
         "reinit",
+
+        #
+        # Built-in Array Functions
+        #
+        "promote",
+
+        # Dimension and Size Functions
+        "ndims",
+        "size",
+        # Dimensionality Conversion Functions
+        "scalar",
+        "vector",
+        "matrix",
+        # Specialized Array Constructor Functions
+        "identity",
+        "diagonal",
+        "zeros",
+        "ones",
+        "fill",
+        "linspace",
+        # Reduction Functions and Operators
+        "min",
+        "max",
+        "sum",
+        "product",
+        # Matrix and Vector Algebra Functions
+        "transpose",
+        "outerProduct",
+        "symmetric",
+        "cross",
+        "skew",
+        # TODO: Where does this one belong?
+        "cat",
     }
 
     def __init__(self, *args, **kwargs):
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -853,6 +853,7 @@ class Class(Node):
                         raise NonConstantSymbolFoundError(f"Symbol '{component_ref}' found in enclosing scope, but is not a constant.")                   
             except KeyError:
                 # Try imports
+                # TODO: Prettyify, bit of duplicate code between if and elif branches
                 if "*" in self.imports:
                     for package_ref in self.imports["*"].components:
                         imported_comp_ref_class = package_ref
@@ -865,6 +866,16 @@ class Class(Node):
                                 raise NonConstantSymbolFoundError(f"Symbol '{component_ref}' found in imports, but is not a constant.")
                         except (KeyError, ClassNotFoundError):
                             pass
+                elif component_ref.name in self.imports:
+                    # TODO: Probably need some test cases to cover the
+                    # ImportClause etc stuff in the similar code in find_class
+                    # and implement the same logic here if needed (not sure yet, it being in find_class might be good enough).
+                    sym = self.find_symbol(self.imports[component_ref.name])
+                    if "constant" in sym.prefixes:
+                        return sym
+                    else:
+                        raise NonConstantSymbolFoundError(
+                            f"Symbol '{component_ref}' found in imports, but is not a constant.")
 
                 if search_parent and self.parent is not None:
                     return self.parent._find_symbol(component_ref)
--- a/src/pymoca/ast.py
+++ b/src/pymoca/ast.py
@@ -19,7 +19,19 @@ class ConstantSymbolNotFoundError(Exception):
     pass
 
 
+class SymbolNotFoundError(Exception):
+    pass
+
+
+class NonConstantSymbolFoundError(Exception):
+    pass
+
+
 class FoundElementaryClassError(Exception):
+    pass
+
+
+class UnsupportedFullReferenceLookup(Exception):
     pass
 
 
@@ -153,6 +165,7 @@ class ComponentRef(Node):
         self.name = ""  # type: str
         self.indices = [[None]]  # type: List[List[Union[Expression, Slice, Primary, ComponentRef]]]
         self.child = []  # type: List[ComponentRef]
+        self._resolved_symbol = None  # type: SymbolReference
         super().__init__(**kwargs)
 
     def __repr__(self) -> str:
@@ -222,6 +235,23 @@ class ComponentRef(Node):
         n.child = [b]
         return a
 
+
+class AttributeRef(ComponentRef):
+    def __init__(self, component_ref: ComponentRef):
+        assert len(component_ref.child) == 0
+        assert component_ref.name in Symbol.ATTRIBUTES
+        assert component_ref.indices == [[None]]
+
+        super().__init__(name=component_ref.name)
+
+
+class SymbolTypeRef(ComponentRef):
+    def __init__(self, component_ref: ComponentRef):
+        assert len(component_ref.child) == 0
+        assert component_ref.name == "Real"
+        assert component_ref.indices == [[None]]
+
+        super().__init__(name=component_ref.name)
 
 class Expression(Node):
     def __init__(self, **kwargs):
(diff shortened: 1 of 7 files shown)


jackvreeken:
So @kentrutan , the issue is definitely in the flattener. Given the simplest-offending case made by @SGeevers :

```
package Package
  type Length = Real;
  constant Real KM_TO_M = 1000.0;
end Package;

model Model
  constant Real KM_TO_M = 1000;
  Package.Length distance_km = 1.0;
  Package.Length distance_m = Package.KM_TO_M * distance_km;
end Model;
```
OMC gives

```Modelica
class Model
  constant Real KM_TO_M = 1000.0;
  Real distance_km = 1.0;
  Real distance_m = 1000.0 * distance_km;
end Model;
```

Now, we do things a bit different in Pymoca, but we should end up with

```Modelica
class Model
  constant Real KM_TO_M = 1000.0;
  Real distance_km = 1.0;
  Real Package.Length = 1000.0;
  Real distance_m = Package.Length * distance_km;
end Model;
```

In other words, we keep the constants around, and just add them with their flattened name in the eventual model. I think we needed this for RTC-Tools at some point, but it's non-compliant behavior I guess (although possibly still desired).

Now, where things get funky is the `ConstantReferenceApplier`. It correctly finds the constant reference, correctly flattens it, but it adds it to the wrong class. It's _supposed_ to add it to the Model class, but instead it adds it to the `Length` InstanceClass, which is somewhere part of the _type_ of the `distance_m` symbol.

So what happens in `ConstantReferenceApplier` is:
1. Entering Model
2. Entering Length 
3. Finding reference to constant -> storing
4. Exiting Length - popping off the stored constant references
5. Exiting Model

I obviously didn't expect to find InstanceClasses other than those relevant to scope of variables, so what I expected when I made the code is:
1. Entering Model
2. Finding reference to constant -> storing
3. Exiting Model

The fix that I made now feels ugly, And I'm not even sure if it's correct in all cases either... Can you have a constant reference in an InstanceClass of a `__builtin` type? All this flattening stuff is currently loaded in your brain-RAM, so hopefully you have a better idea

kentrutan:
> The fix that I made now feels ugly, And I'm not even sure if it's correct in all cases either... 

I am also a little suspicious. Since I'm working on the earlier stuff, I have spent little to no time looking at `apply_constant_references()` and am not super sure what it is supposed to do. 

When I looked at the point of failure in the generator, the problem looked to me like it was caused by not using Modelica name lookup rules, which sounds consistent with some of your discussion above. 

I will step through it to see if I can spot anything, but it will likely be several days with other stuff I have going on. I'm going to mark this draft for now.

jackvreeken:
Three tests still fail because they have modifications _without_ a scope set. I'll have to dive in to see why and what that means. 

Aside from that:
- Still not sure using `scope` to add symbols to is the correct approach
- The `ConstantReferences.mo` test file fails to flatten with OMC 🤷‍♂️

kentrutan:
> * The `ConstantReferences.mo` test file fails to flatten with OMC with some non-descriptive error 🤷‍♂️

This fixes it for omc:

```diff
diff --git a/test/models/ConstantReferences.mo b/test/models/ConstantReferences.mo
index ffa5ad9..dfbdf2c 100644
--- a/test/models/ConstantReferences.mo
+++ b/test/models/ConstantReferences.mo
@@ -1,3 +1,4 @@
+package P
 package P0
   constant Real p[2] = {1, 2};
 end P0;
@@ -27,7 +28,7 @@ model b
   extends a(redeclare package m = m2);
   Real x;
   parameter Real y = m.p;
-  parameter Real z = P0.p[0];
+  parameter Real z = P0.p[1];
   Real w;
   Real g;
 equation
@@ -35,3 +36,4 @@ equation
   x = m.p;
   w = m.p * 2.0;
 end b;
+end P;
```

jackvreeken:
I'm going down the rabbit hole of the flattening process chasing this bug. I expect to eventually run into a blocking issue that requires a major overhaul, i.e. your reworking of it all. Until that point though, I'm refreshing in my mind how the code/flattening process works; it's been a few years after all😅 Hopefully then I can also be more helpful for e.g. reviewing your future changes to this flattening process, or helping out when issues arise.

kentrutan:
I was wondering if you might run into the same dead end as I met. Unfortunately, my availability to follow your work has been about nil and will be for probably the rest of this week. Sorry.

kentrutan:
Your latest changes are failing `test_constant_references` and `test_parameter_modification_scope`.  Neither appear to be failing in areas related to `extends` scope that I am working on for #266. Two added tests I'm using for #266 also both fail with your latest code, I guess not surprisingly.

I am working on loss of scope for symbol types coming from `extends` statements, you are working on scope of constant symbols, but they are both having problems in the instantiation and flattening method in Pymoca. If you were to fix scoping of constants, I believe it will still fail in cases like #266 that contain constants because the scope is lost in an earlier step in the instantiation/flattening. It is possible that these are two separate issues, but it is also highly possible they are related. 

You are welcome to continue plugging away at this one to continue to refresh you memory and perhaps fix all existing test cases, but I'm also OK if you want to take a pause to wait for my work. I do expect to ramp up back up on it starting this week.

kentrutan:
By the way, my patch above for `ConstantReferences.mo` was a bit bigger than needed to pass OMC. This is all that is needed:
```diff
diff --git a/test/models/ConstantReferences.mo b/test/models/ConstantReferences.mo
index ffa5ad9..07d9c60 100644
--- a/test/models/ConstantReferences.mo
+++ b/test/models/ConstantReferences.mo
@@ -27,7 +27,7 @@ model b
   extends a(redeclare package m = m2);
   Real x;
   parameter Real y = m.p;
-  parameter Real z = P0.p[0];
+  parameter Real z = P0.p[1];
   Real w;
   Real g;
 equation
```
In other words, Modelica array indexes start at 1, not 0. 

The other part of the earlier patch was intended to make it compatible with spec section 13.2.2.2, but that change is only needed if loaded from MODELICAPATH. OMC is OK with just the above when directly loading the file.

jackvreeken:
> You are welcome to continue plugging away at this one to continue to refresh you memory and perhaps fix all existing test cases, but I'm also OK if you want to take a pause to wait for my work. I do expect to ramp up back up on it starting this week.

I'll probably keep plugging away at it, and will also see if my new approach helps/hurts the extends stuff as well (it's all somewhat related obviously). I do obviously want to avoid duplicate work between the two of us, but I also want to understand the problem space a bit more. And, seeing as how we've been cut many times by this whole flattening/scoping before, I want to avoid polluting your mind with whatever ideas I have now on how to possibly do things as I have been wrong before 😅
 
Maybe it could be a good idea to both push things a little further, but sit together to discuss well before we reach the "polishing up" phase? What do you think?

kentrutan:
> Maybe it could be a good idea to both push things a little further, but sit together to discuss well before we reach the "polishing up" phase? What do you think?

Sounds good. Glad you are pushing it. I'm going to refrain from looking at your code while the commits are flying. Let me know when you want me to take a look.

I'll create a draft PR for what I'm doing and start with commits for the two added test cases so you can easily grab them.

jackvreeken:
@kentrutan I'm starting to think more confidently about this new approach. Still many (MANY) things to do, reason about, fix, or prettify, but overall I think we stand to gain a lot. For example, we can actually raise proper exceptions when people refer to symbols that don't exist, or exist outside the current scope but are not constant, or have already been declared. Currently we're just hoping for the best, and it's e.g. the CasADi backend that raises "missing symbol" errors. I also think it's conceptually also a lot easier to understand the whole flattening process and to fix bugs in it.

Before putting any effort into more than just testing this new concept, I want to check whether you have maybe built something similar already (and that I can now actually understand after loading all this in my brain-RAM again). Were you working solely on the extends thing (i.e. modifying mostly "build_instance_tree" and "flatten_extends"), or basically also an overhaul of the entire flattening process including `modify_symbols` / constant ref flattening / constant pulling / function pulling / etc?

kentrutan:
I started with a focus on the extends scope issue, but as I study the Modelica spec closely, I see more areas that Pymoca does not match, so it is moving more toward a larger rewrite. Certainly name lookup, instantiation, and flattening are changing. I would say it is going to take at least a month, No Promises™️. I hopefully will be confident enough to share some notes and structure before that and that point we can compare high level implementations. I think I'll refrain from looking at your code until I have something to share. Alternately, if you think you are ready to release and I can try to poke holes in it.

I really like the idea of having proper exceptions and also pulling the generic stuff out of the CasADi backend. Good stuff!

Have you tested your code with the two test cases that I put into #307?

jackvreeken:
If you're still in the "making high level design" phase then it's all good, I desperately want to avoid either of us implementing something merge worthy when that might be wasted effort. But separately coming up with high level designs (and obviously some hacky prototyping to test) is exactly what I think we need, and then we can combine forces and discuss how our approaches might differ in some of the (very weird) cases that Modelica supports.

I have my high level design somewhat ready at this point. I'll push it a little further this weekend to see how I do on your test cases, and also polish up the bunch test cases I made, amended and/or copied.

src/pymoca/ast.py
```diff
@@ -756,6 +833,76 @@ def _find_constant_symbol(self, component_ref: ComponentRef, search_parent=True)
     def find_constant_symbol(self, component_ref: ComponentRef) -> Symbol:
         return self._find_constant_symbol(component_ref)
 
+    def _find_symbol(self, component_ref: ComponentRef, search_parent=True, in_original_scope=False) -> Symbol:
+        if component_ref.child:
+            # Try classes first, and constant symbols second
+            t = component_ref.to_tuple()
+
+            try:
+                node = self._find_class(ComponentRef(name=t[0]), search_parent)
```
jackvreeken, line 842: Probably faster to make this a `copy=False` lookup. That's also strictly speaking more correct.

src/pymoca/ast.py
```diff
@@ -756,6 +833,76 @@ def _find_constant_symbol(self, component_ref: ComponentRef, search_parent=True)
...
+                try:
+                    s = self.symbols[t[0]]
+                except KeyError:
+                    raise SymbolNotFoundError()
+
+                # Found a symbol. Continue lookup on type of this symbol.
+                if isinstance(s.type, InstanceClass):
+                    return s.type._find_symbol(ComponentRef.from_tuple(t[1:]), False, in_original_scope)
+                elif isinstance(s.type, ComponentRef):
+                    node = self._find_class(s.type)  # Parent lookups is OK here.
```
jackvreeken, line 854: Same here, Probably faster to make this a copy=False lookup. That's also strictly speaking more correct.

test/models/ExtendsOrder.mo
```diff
@@ -21,14 +23,25 @@ package P
...
+        parameter Real n(nominal=bla, min=P.min_m, max=2 * max_m) = 0;
       equation
-        x = m;
+        x = m * max_m;
       end BT;
-      BT bt;
+      BT bt(bla=150);
     end B;
 
     extends A(redeclare model AT = BT); // Here we expect to find P.M.B.BT, not P.BT
```
jackvreeken, line 36: I think the way to do this is to shift all extends modifiers first to all "partially instantiated" classes, and only then extend. Because at least OMC applies the modifier for BT(bla=200) also on the BT in `extends A(redeclare model AT=BT)` does

kentrutan:
I see that you are making some substantial changes, including name lookup. My current work is on a unified name lookup conforming to the spec. It's only half-baked, so I don't want to waste your time by pushing commits. At a high level, I'm toying with moving it from `ast.Class` to tree.py, since name lookup can also return an `ast.Symbol`. In addition, MODELICAPATH uses some similar lookup logic, so I'm seeing how DRY I can make it.

src/pymoca/ast.py
```diff
@@ -756,6 +833,76 @@ def _find_constant_symbol(self, component_ref: ComponentRef, search_parent=True)
...
+                else:
+                    # Not in the original scope, and then we are only allowed to refer to constants.
+                    if "constant" in sym.prefixes:
+                        return sym
+                    else:
+                        raise NonConstantSymbolFoundError(f"Symbol '{component_ref}' found in enclosing scope, but is not a constant.")                   
+            except KeyError:
+                # Try imports
+                # TODO: Prettyify, bit of duplicate code between if and elif branches
+                if "*" in self.imports:
```
kentrutan, line 872: I also just ran across this ugly import code I wrote while working on name lookup. It works, but not so easy to follow, even for me. I'm looking at cleaning it up some, starting in the parser. Just wanted you to know that I'm working on your TODO.

jackvreeken:
> I see that you are making some substantial changes, including name lookup. My current work is on a unified name lookup conforming to the spec. It's only half-baked, so I don't want to waste your time by pushing commits. At a high level, I'm toying with moving it from `ast.Class` to tree.py, since name lookup can also return an `ast.Symbol`. In addition, MODELICAPATH uses some similar lookup logic, so I'm seeing how DRY I can make it.

That'd be great! A combined symbol/class lookup makes a lot of sense, I just duplicated code for this proof-of-concept. I don't think I'll make many more changes until you have something to show me, and definitely not clean ones. I _might_ play around with the extends/redeclarations a bit for your test cases, but that's it.

Since you're already looking at my code, the high level design changes are basically what's in `def flatten_class`. Really high over the two most important changes are:
- adding `resolve_component_references_to_symbols`. I do this in a step _after_ the building of the instance tree, where I loop over all ComponentRefs. I'm basically just interested in the ones to functions and symbols, and link them to a ast.Symbol class (like, the actual Python object) or an ast.InstanceClass for functions. 
- simplifying `flatten_component_refs` . Because we have linked ComponentRefs to ast.Symbols, we don't have to keep track of _every_ ComponentRef (e.g. those inside equations) while flattening. We just have to keep track of the name of `Symbols` part of `InstanceClass.symbols` (and functions again, which is a TODO). After we're done flattening the symbols, we just loop over all ComponentRefs (those in equations, nominals, etc), ask which Symbol they point to, and then we know what the new flattened name is (as a string). Done. No more "copy expression" stuff that we had before, no more updating _every_ component ref at every flattening step without knowing if the symbol even exists.

The `resolve_component_references_to_symbols` step adds this "symbol lookup" that you're also implementing (probably better than my hacky thing), and can do all the relevant checks of "symbol found but outside scope and not constant" (or just even "symbol not found").

Aside from that there's a bit of scoping of modifications, and some other code that was made redundant. And many more things that _should_ be changed to work nicely with the above way of doing things.

Some more notes:
- I think storing the "_resolved_symbol" in ComponentRef is ugly, but it worked as a proof-of-concept
- Everything being a ComponentRef is annoying; some are to classes, some to symbols, some to functions, some are import statements, and some are names of attributes (like nominal, value, min, etc). Definitely not all created equal, and would be nice to tell them apart more easily (`isinstance(tree, ComponentRefToSymbol)` or something)

kentrutan:
Thanks for the high level overview on your changes. Super helpful! With that, I'll take a closer look. I haven't really been looking at it recently, just responding when I see keywords in emails related to my work.

> I might play around with the extends/redeclarations a bit for your test cases, but that's it.

Please do! I am interested to see if what you have is on the right track and then that might influence what I'm doing (like throw it out 😆).

I still think it is good that we work on separate forks for a while longer before combining forces. As you said, let's get together and discuss, but I am really not ready yet. Clearly you crank out functioning code faster than me. 🥇

kentrutan:
@jackvreeken I started looking at your latest code here and see that there are a lot of tests failing at the current HEAD, but that's the commit where you added the two tests from #307. Is there a particular commit I should look at for the latest/best working version?

kentrutan:
@jackvreeken Now that parse cache CI on Windows is fixed, wondering if you have time for some instantiation/flattening work? See my previous comment here.

jackvreeken:
Yes, it was my plan to dive back into this MR to refresh my memory. It's still very PoC'ish though, and definitely just an inspiration for whatever final solution we come up with. Not sure I'll get around to it tomorrow, otherwise it's the weekend.

Noted on the failing tests, I'll start then by seeing if everything works without them. And then the big question I stopped at last time, is why those two tests are failing. What exactly is happening in flatten_extends that conflicts with the spec/expected output.

kentrutan:
> Not sure I'll get around to it tomorrow, otherwise it's the weekend.

Thanks for letting me know. Don't stress out about getting it done even this weekend. I have more homework to do on implementation plans before being able to share and I have also have other work to do the rest of this week. Just wanted to see it's on your radar. It will be great to work together on it - your insight is invaluable! Cheers.

jackvreeken:
This PR was a playground for a proper implementation (see #307  )

PR #308, opened 2023-10-08
Update .git-blame-ignore-revs

kentrutan:
The previous version of this file had commit hashes from the PR branch. This version uses the merged commit hashes on master and adds another commit from #289 that is only about formatting.

Lesson learned: This will have to be updated after any formatting-only commits in a PR are merged into master/main.

EDIT:
Prior to this change, I noticed git blame was referencing formatting-only changes when I was trying to look at recent pull request changes.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -18,13 +18,17 @@
 # - Add a link to the relevant issue/MR
 
 # START https://github.com/pymoca/pymoca/pull/281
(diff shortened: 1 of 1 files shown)


jackvreeken:
Hmm, interesting indeed. I would've thought that I rebased properly on top of everything before merging. Next time around, will make sure to do the .git-blame-ignore-revs update in a separate MR, to avoid accidentally (or always?) having the wrong hashes.

jackvreeken (approved):
LGTM

PR #309, opened 2023-10-08
CI/CD: Update max Python to 3.12

kentrutan:
Try Python 3.12 in the test matrix.

Also since there are so many versions of Python active (3.8 to 3.12), it seems excessive to test every version. This changes the matrix to just three versions: 3.8, 3.10, 3.12. For future updates, I envision picking min, max, and median of the supported Python versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: "3.12"
       - uses: pre-commit/action@v3.0.0
 
   test:
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.10", "3.12"]
 
     steps:
     - uses: actions/checkout@v3
@@ -54,7 +54,7 @@ jobs:
         submodules: true
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.11"
+        python-version: "3.12"
     - run: |
         python -m pip install --upgrade pip
         pip install tox
@@ -72,7 +72,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.11"
+        python-version: "3.12"
     - run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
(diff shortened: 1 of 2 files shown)


kentrutan:
The casadi wheel build is failing for Python 3.12, so I guess we'll need to wait and try this again at a later date.

jackvreeken:
@kentrutan CasADi released 3.6.4 with Python 3.12 wheels two days ago, so this MR is now ready.

I made some changes to basically just restart the CI/CD pipelines, I'll leave it to you to squash/properly rebase on top of master.

kentrutan:
@jackvreeken Because of these delays, we should probably pin max Python version until all dependencies support the new Python version. What do you think?

At some point, we will want to pin versions of all dependencies. What do you think of using a tool like hatch or poetry to help with this? Not saying we move to that now, but maybe at the same time we are ready to migrate more/all of setup.py into latest [recommended packaging practice](https://packaging.python.org/en/latest/).

jackvreeken:
I think I fixed this one, it does depend on #313 and #314 , so those need merging first

kentrutan:
@jackvreeken What fixed the problem: getting rid of setup.py, the CasADi updates, or both?

Do you have any thoughts on my [earlier comment](https://github.com/pymoca/pymoca/pull/309#issuecomment-1804139796) now that we (you) have some more experience with the issue?

> @jackvreeken Because of these delays, we should probably pin max Python version until all dependencies support the new Python version. What do you think?
> 
> At some point, we will want to pin versions of all dependencies. What do you think of using a tool like hatch or poetry to help with this? Not saying we move to that now, but maybe at the same time we are ready to migrate more/all of setup.py into latest [recommended packaging practice](https://packaging.python.org/en/latest/).

If you agree, it might be a good time to pin max Python with this PR.

jackvreeken:
Fixing 3.12 was:
- Waiting for CasADi to have packages on PyPI
- Fixing distutils deprecation and adding setuptools as a dependency (for the casadi backend). Note that we can then still use import `distutils.ccompiler`, but we _have_ to have setuptools installed.

--- 

Hmm, my thoughts on pinning (exact versions) is that: 
- cli tools/applications probably should
- libraries should probably not

So for most dependencies we can use semver and pin the major version in `major.minor.patch`; for CasADi we have to pin `major.minor.*`. And limiting Python to 3.12 is also fine by me, but then I do think in general we should have pipelines running for the latest version we _haven't_ pinned. Renovate was one of these tools we could use, but there are definitely others. I also spotted something in a `.pre-commit-config` file the other day about updating those (black/flake8) with some interval of like a month.

kentrutan:
OK I'll pin some versions per your [comment](https://github.com/pymoca/pymoca/pull/309#issuecomment-1913160144). PR #315 (next to be merged) looks like a better place to do it.

PR #313, opened 2024-01-24
CI/CD: Fix cache test

jackvreeken:
When running on Github Actions, the first and second hit time have a tendency to be equal, possibly for reasons related to mounted/network storage.

To make sure that the cache is hit, we can also check whether the log messages indicate as such, instead of just looking at a forced update of the last hit time.

Note that this commit does not fix the failing of the test, it just adds an additional check to better understand what is going on.

### DIFF
--- a/test/parse_test.py
+++ b/test/parse_test.py
@@ -785,8 +785,14 @@ class ParseTest(unittest.TestCase):
             cursor.execute("SELECT COUNT(*) FROM models")
             self.assertEqual(cursor.fetchone()[0], 1)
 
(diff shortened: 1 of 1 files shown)


kentrutan (approved):
This is surely a tricky one to debug! I don't have anything to add other than just looking a the code. It looks OK to me.

kentrutan:
Sorry to say it's still failing.

kentrutan:
Reran Windows and it passed. Another clue is that so far it has passed on second tries.

PR #314, opened 2024-01-24
Switch from setup.py to pyproject.toml

jackvreeken:
- [x] Merge #313 first


kentrutan (approved):
I pushed a couple commits with my suggested changes. Thought it would be easier than review comments.

kentrutan:
Second rerun of failed CI didn't work this time. I'll try a 3rd.

kentrutan:
Third time's a charm for the CI. 

@jackvreeken let me know if you are OK with my changes and I will squash the "cruft" commit and merge.

jackvreeken:
Looks good to me! I actually prefer the separate commits (so the 3 we have right now), because then it's easy to see the changes. If we _both_ move to a new file _and_ change lines, it's harder to diff. So as far as I'm concerned you can go ahead and merge the PR as-is, assuming you have other changes you want to make.

PR #318, opened 2024-01-29
Cache timing cleanups and improvements

jackvreeken:
- [x] Needs #317

### DIFF
--- a/src/pymoca/parser.py
+++ b/src/pymoca/parser.py
@@ -11,8 +11,9 @@ import os
 import pickle
 import platform
 import sqlite3
+import time
 from collections import OrderedDict, deque
-from datetime import datetime, timedelta
+from datetime import timedelta
 from pathlib import Path
 from typing import Dict, List, Optional, Union  # noqa: F401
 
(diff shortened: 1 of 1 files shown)


src/pymoca/parser.py
```diff
@@ -1033,12 +1035,13 @@ def parse(
...
-            (datetime.utcnow() - timedelta(days=cache_expiration_days)).timestamp() * 1e6
-        )
+        cutoff_time = _microseconds_since_epoch(timedelta(days=-cache_expiration_days))
         cursor.execute("DELETE FROM models WHERE last_hit < ?", (cutoff_time,))
+        # Sometimes Windows time resolution is a bit coarse, so we make
+        # sure that if we update the last_prune time, it is actually newer
+        # than the previous one.
         cursor.execute(
-            "UPDATE metadata SET value = ? WHERE key = ?",
+            "UPDATE metadata SET value = max(value + 1, ?) + 1 WHERE key = ?",
```
kentrutan, line 1044: Comparing this to line 1080, there is an extra `+ 1` before the `WHERE`. Is that what you want?

kentrutan (approved):
I explain on line 1044 the +2 comment I had earlier.

src/pymoca/parser.py
jackvreeken: Ahh, that's a good catch! That +1 at the end is a remnant of me trying `? + 1` first, before realizing I could just do the proper thing with `max()`.

PR #320, opened 2024-02-20
CasADi backend:  Special-case binary operations on elements with numel = 0

jbaayen:
Since CasADi 3.6, two matrices with zero elements, but with different dimensions (i.e., 0x1, 1x0, or 0x0), can no longer be added or subtracted. This change checks for lhs and rhs having zero elements before adding or subtracting in CasADi.


jackvreeken:
Can you add a small test case? I want to avoid accidentally breaking this again in the future, as this stuff is easy to overlook/forget.

jbaayen:
Absolutely. I've added one now.

jackvreeken:
The test runs fine on master, so doesn't actually test the fixes?

jbaayen:
The test doesn't run fine on 0.9.x, so perhaps we only need the fix on the maintenance branch.

jbaayen:
I'm closing this as I can no longer reproduce the issue on 0.9.x either.

Issue #316, opened 2024-01-26
Flattening: Arrays of models containing scalar equations generate scalar equations acting on arrays


jbaayen:
Flattening a model that contains arrays of instances of model components, that contain scalar equations, fails. The scalar equations get flattened to scalar equations acting on arrays, which then triggers a CasADi error:

```
RuntimeError: .../casadi/core/sparsity.cpp:414: Assertion "x.size2()==y.size1()" failed:
Matrix product with incompatible dimensions. Lhs is 2x1 and rhs is 2x1.
```

Example code: 

```
import os

from pymoca.backends.casadi.api import transfer_model

mo_content = """
model Component
  input Real x;
  input Real y;
  output Real z;
equation
  z = x * y;
end Component;

model Test
  parameter Integer dim = {};
  input Real ix[dim];
  input Real iy[dim];
  Component c[dim](x = ix, y = iy);
end Test;
"""

try:
    os.mkdir("flatten-arrays-test")
except:
    pass

for dim in [1, 2]:
    print("Testing dim={}".format(dim))
    with open("flatten-arrays-test/tmp.mo", "w") as f:
        f.write(mo_content.format(dim))

    transfer_model("flatten-arrays-test", "Test")
    print("OK")
```

jackvreeken:
Apparently there's a workaround available, so I suggest we keep chugging away at #306 / #307 first

jackvreeken:
@jbaayen You are aware that your example fails in OMC?

`omc -i=Test .\issues\issue_316.mo` gives

```
rror processing file: .\issues\issue_316.mo
Failed to parse file: .\issues\issue_316.mo!

[C:/dev/pymoca/issues/issue_316.mo:10:27-10:28:writable] Error: Parse error: Empty array constructors are not valid in Modelica.

# Error encountered! Exiting...
# Please check the error message and the flags.
Failed to parse file: .\issues\issue_316.mo!
```


I didn't look too much further, but a test case that fails in Pymoca, but works/flattens as expected in OMC is easier to work with.

kentrutan:
Haha I did the same thing. You need to run the whole script he provided to show the error.

jackvreeken:
You're right lol. I even tried with `dim = {1}` with my tired head, not realizing that I was doing the substitution wrong.

jbaayen:
[...]

PR #321, opened 2024-02-22
Keep track of the depth at which a symbol is nested in arrays

jbaayen:
The sum() operator acts along the "most nested" dimension.  To make sure this maps to the correct CasADi operation, we must keep track of how deeply a symbol has been nested in arrays.


jbaayen:
This is a fix for the 2nd test case in #316.

jbaayen:
I'm not sure I understand the first example. What is the role of `v_sum`?

PR #323, opened 2024-04-03
Fixes for 0.9.1 -> 0.9.2 

jackvreeken:
A string comparison for versions would result in "3.10.0" being smaller than "3.5.0" (because 3.1 is smaller than 3.5). We should just do a tuple comparison.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,8 @@ ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
 # pylint: disable=no-init, too-few-public-methods
 
 
-PYTHON_VERSION = '.'.join([str(i) for i in sys.version_info[:3]])
-PYTHON_VERSION_REQUIRED = '3.5.0'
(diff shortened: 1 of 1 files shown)


kentrutan:
Wondering why you didn't just use python_requires like 1cec443? Minimum risk change for a maintenance branch?

kentrutan:
Stated another way, can you use the patch to setup.py from 1cec443 but keeping version>=3.5? I may be missing something.

kentrutan:
Ahh so there *is* a slight possibility of a problem using `python_requires`. Good catch! But it looks like there is a problem with the CI test on Windows with Python 3.5 so is it going to be a challenge supporting 3.5 even on a maintenance branch?

kentrutan:
Restarting the test to see if macOS Python 3.5 passes (cancelled because Windows failed).

jackvreeken:
Continued in #345

PR #325, opened 2024-05-09
control.rlocus klist 

gdmcbain:
Fixes #324

Issue #324, opened 2024-05-09
control.rlocus kwarg klist->gains


gdmcbain:
https://github.com/pymoca/pymoca/blob/9d597d76aa2b82bd3d563d4938a098af8baafd49/test/notebooks/Spring.ipynb#L1265

raises 

```txt
TypeError                                 Traceback (most recent call last)
Cell In[29], [line 1](vscode-notebook-cell:?execution_count=29&line=1)
----> [1](vscode-notebook-cell:?execution_count=29&line=1) control.rlocus(ss[0,0], klist=pl.logspace(-2,1,1000));
      [2](vscode-notebook-cell:?execution_count=29&line=2) pl.grid()

File c:\Users\G.McBain\.venv\pymoca\Lib\site-packages\control\rlocus.py:188, in root_locus_plot(sysdata, kvect, grid, plot, **kwargs)
    [185](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/rlocus.py:185)     return responses.loci, responses.gains
    [187](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/rlocus.py:187) # Plot the root loci
--> [188](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/rlocus.py:188) out = responses.plot(grid=grid, **kwargs)
    [190](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/rlocus.py:190) # Legacy processing: return locations of poles and zeros as a tuple
    [191](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/rlocus.py:191) if plot is True:

File c:\Users\G.McBain\.venv\pymoca\Lib\site-packages\control\pzmap.py:127, in PoleZeroData.plot(self, *args, **kwargs)
    [123](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/pzmap.py:123)     kwargs = kwargs.copy()
    [124](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/pzmap.py:124)     kwargs['grid'] = config._get_param(
    [125](file:///C:/Users/G.McBain/.venv/pymoca/Lib/site-packages/control/pzmap.py:125)         'rlocus', 'grid', kwargs.get('grid', None), _rlocus_defaults)
[...]

PR #327, opened 2024-05-10
rlocus klist->gains 

gdmcbain:
Replaces #325 as a fix for #324.

### DIFF
--- a/test/notebooks/Spring.ipynb
+++ b/test/notebooks/Spring.ipynb
@@ -1262,7 +1262,7 @@
     }
    ],
    "source": [
-    "control.rlocus(ss[0,0], klist=pl.logspace(-2,1,1000));\n",
+    "control.rlocus(ss[0,0], gains=pl.logspace(-2,1,1000));\n",
     "pl.grid()"
    ]
   },


jackvreeken:
Does this affect the minimum version of `control` required for running the examples? If so, could you add it to pyproject.toml (under the examples section).

If I understand correctly, it's `klist` -> `kvect` -> `gains`? And we're jumping over the `kvect` change straight to `gains`?

gdmcbain:
> Does this affect the minimum version of `control` required for running the examples? If so, could you add it to pyproject.toml (under the examples section).

`control` doesn't appear in `pyproject.toml`.  It does in https://github.com/pymoca/pymoca/blob/9d597d76aa2b82bd3d563d4938a098af8baafd49/tools/create_conda_env.sh#L26

When I installed `pymoca` and ran `Spring.ipynb`, I didn't have `control` at all but on getting

> ModuleNotFoundError: No module named 'control'

I issued

```python
uv pip install control
```

and then hit #324; hence worked on #325 and this.

But  you're right, `gains` seems to have been introduced in [c638362](https://github.com/python-control/python-control/blame/c6383624b390c617ef8a2f0818cfb6994961bb30/control/rlocus.py) on 2024-04-04 but not yet released.  [PyPI](https://pypi.org/project/control) currently has [0.10.0](https://github.com/python-control/python-control/releases/tag/0.10.0), in which [`control.rlocus`](https://github.com/python-control/python-control/blob/0.10.0/control/rlocus.py) has

[`rlocus = root_locus_plot`](https://github.com/python-control/python-control/blob/1b2abe39ea8fa2ac152e4a53fdcb7c3d32e718b9/control/rlocus.py#L450)

and 

[`def root_locus_plot(
        sysdata, kvect=None, grid=None, plot=None, **kwargs):`](https://github.com/python-control/python-control/blob/1b2abe39ea8fa2ac152e4a53fdcb7c3d32e718b9/control/rlocus.py#L100-L101) 

and 

[`        responses = root_locus_map(sysdata, gains=kvect)`](https://github.com/python-control/python-control/blob/1b2abe39ea8fa2ac152e4a53fdcb7c3d32e718b9/control/rlocus.py#L169)

So I guess it would be best to use `kvect` here and pin `control==0.10.0`.  I'll close this and prepare a new PR.

PR #330, opened 2024-05-28
CI/GA: Update actions to their latest versions

jackvreeken:
GitHub Actions was giving deprecation warnings about NodeJS 16, and that we should update to NodeJS 20. We therefore update all actions to their latest versions.


kentrutan:
I check that the versions are still the latest. Looks good and seems to work fine. 

Great to see those Codecov reports again with the new token!

Closes #332

PR #339, opened 2024-12-04
codegen: Workaround MSVC NaN constantness bug

### LINKED ISSUE #338: CasADi: CI fail on Windows in gen_casadi_test.py `test_codegen`
In CI on the Windows environments, we are getting the following compile errors in a file generated by CasADi 3.6.7. This version of CasADi was released Sept. 30, 2024 and there were successful Windows CI runs with this version until the first failure Nov. 20. There was a change in the Windows image between these dates.

```
test\models\Aircraft_variable_metadata.c(144): error C2099: initializer is not a constant
test\models\Aircraft_variable_metadata.c(145): error C2099: initializer is not a constant
```

These are the offending lines of Aircraft_variable_metadata.c generated on macOS:

```C
static const casadi_real casadi_c0[6] = {casadi_nan, casadi_nan, -casadi_inf, -casadi_inf, casadi_inf, casadi_inf};
static const casadi_real casadi_c1[18] = {casadi_nan, casadi_nan, casadi_nan, casadi_nan, casadi_nan, casadi_nan, -casadi_inf, -casadi_inf, -casadi_inf, -casadi_inf, -casadi_inf, -casadi_inf, casadi_inf, casadi_inf, casadi_inf, casadi_inf, casadi_inf, casadi_inf};
```

I don't have a Windows machine to debug this.

jackvreeken:
Version 10.0.26100 of the Windows SDK introduced changes to how INFINITY and NAN are defined in `math.h`. This resulted in NAN no longer being a compile-time constant, and codegen'd code failing compilation the most recent `windows-latest` image. The workaround is to define _UCRT_NOISY_NAN, which restores the old behavior.

Closes #338

### DIFF
--- a/src/pymoca/backends/casadi/api.py
+++ b/src/pymoca/backends/casadi/api.py
@@ -145,7 +145,9 @@ def _codegen_model(model_folder: str, f: ca.Function, library_name: str):
         compiler_flags = ["-O2", "-fPIC"]
         linker_flags = ["-fPIC"]
     else:
-        compiler_flags = ["/O2", "/wd4101"]  # Shut up unused local variable warnings.
+        # Shut up unused local variable warnings, and use workaround for
+        # NaN-constantness bug in recent Windows SDK versions.
+        compiler_flags = ["/O2", "/wd4101", "/D_UCRT_NOISY_NAN"]
         linker_flags = ["/DLL"]
 
     # Generate C code


jackvreeken:
Eventually I hope we can get rid of the workaround, with either CasADi adding the workaround, or Microsoft fixing the bug.

kentrutan (approved):
CI passed, so looks good to me. 

How will we remember to undo this when it is fixed in CasADi or Windows math.h?

jackvreeken:
I made the issue over on the CasADi side, so I will receive updates on it. Not sure how to keep track of the Windows SDK one. We can add a test that assumes compiling _without_ the flag fails on `windows-latest`, but that seems like overkill. Furthermore, even if the `windows-latest` runner might have a correct SDK again in the future, some random user might still have the wrong version installed on their own PC.

Just keeping track of a potential fix over on the CasADi side seems appropriate to me, and then removing the workaround + bumping CasADi's minimum version requirement when it is released.

PR #340, opened 2024-12-27
CI/CD: Update Python >=3.9 and <=3.13

kentrutan:
Tests that ran OK 2 days ago are now failing on Python 3.8 on Windows when there have been no changes to tested code. Since Python 3.8 is end of life, let's bump min Python to 3.9 and max to 3.13 and see how it tests out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: "3.13"
       - uses: pre-commit/action@v3.0.1
 
   test:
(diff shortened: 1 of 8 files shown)


kentrutan:
Tests pass. Looks like it's a go with your approval @jackvreeken.

jackvreeken (approved):
Not sure if the precommit changes were intended or coupled to 3.13, but otherwise approved 👍

kentrutan:
> Not sure if the precommit changes were intended or coupled to 3.13, but otherwise approved 👍

An update to black was needed to support Python 3.13. I decided to go ahead and update flake8 to latest while I was in there.

kentrutan:
Sorry I just realized that I should have split out the formatting only changes into a separate commit so we could add to .git-blame-ignore-revs. Not going to change now that it is merged. Please help me remember to do this @jackvreeken if you notice.

PR #342, opened 2024-12-27
CI/CD: Fix deploy consistent with pyproject.toml

kentrutan:
Change from the old setup.py-based build to something consistent with current Python packaging standards (though the instructions are poorly documented). I guess we have not done a deploy since the change to pyproject.toml, so this is the first time we are seeing the issue.

EDIT: We also fixed some issues with the build and upload to PyPI including the rendering of the README on PyPI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        submodules: true
     - uses: actions/setup-python@v5
       with:
         python-version: "3.13"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,12 +76,12 @@ jobs:
         python-version: "3.13"
     - run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
+        pip install build twine
     - env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
       run: |
-        python setup.py sdist bdist_wheel
+        python -m build
         twine upload dist/*
     - uses: softprops/action-gh-release@v2
       with:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,6 @@
 include LICENSE
 include README.md
 recursive-include doc *
-recursive-include java *
 include src/pymoca/_version.py
 include src/pymoca/Modelica.g4
 include src/pymoca/backends/xml/ModelicaXML/schemas/*.xsd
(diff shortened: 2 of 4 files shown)


kentrutan:
Getting pymoca version 0.0.0. Thought it was versioneer needing Python <= 3.11, but that's not it. Perhaps a setuptools with the new build system is clashing? TBD

kentrutan:
The last push has a solution, but I'm not totally confident that it will work in future versions of setuptools and/or versioneer (details in 173a327 message). Maybe the long-term solution is to look for ways to get #303 to work in the context of #287.

setuptools_scm is currently bit ahead versioneer in the latest release (May 2024 vs July 2023). They are both currently testing only up to Python 3.12, so our support of Python 3.13 is a bit ahead of them, but seems to work OK. Neither test against macOS ☹️.

kentrutan:
Last two commits fix warnings I saw in the build where files from MANIFEST.in were not included in the distribution. Thought I would go ahead and fix for the next release.

kentrutan:
We now have a proper project description on PyPI from the README. There are a couple loose ends related to that that I want to fix for the 0.11.0 release.

- Fix the description in `pyproject.toml` to be consistent with the README (already pushed the commit to fix).
- Fix the broken pymoca image on PyPI that is OK on GitHub README. Yet to be fixed.

kentrutan:
OK I think we are ready for an official 0.11.0 release with all README images and links working on TestPyPI (and hopefully PyPI after the CI/deploy with the 0.11.0.dev6 release completes).

jackvreeken:
Nice cleanups! And for sure, the versioneer + setuptools_scm could do with a revisit. The big thing there is cache invalidation with local development by detecting "dirty" iirc.

kentrutan:
Hello @jackvreeken. Is reviewing this on your radar? No hurry, just wondering.

.github/workflows/ci.yml
```diff
@@ -76,12 +76,12 @@ jobs:
         python-version: "3.13"
     - run: |
```
jackvreeken, line 79: (related to the commit message of the first commit) The paragraph starting with `Adding `setup.py` back in could be a fall-back solution` is not wrapped in the commit message.

jackvreeken (changes requested):
Just some wrapping in the commit message(s). Otherwise looks absolutely great to me!

jackvreeken:
On the topic of pyproject.toml, and what tool to use to build, I've recently been using `uv` a lot at <dayjob>. It's by the makers of Ruff, and it works really smoothly so far. Somewhere down the line I'll give setuptools_scm a go again, maybe now it'll work with your `tool.setuptools.dynamic` section change (and/or `uv` making things easy to install/upkeep)

jackvreeken:
LGTM!

PR #343, opened 2025-01-09
CI/CD: Fix `pymoca.__version__`

kentrutan:
We are currently getting "0+unknown" from the PyPI package. Also, we are getting the setuptools-generated version number in development mode (no "dirty" flag currently needed for the parse cache code).

This fixes it by adding setup.py back in to generate _version.py and removes the setuptools dynamic versioning from pyproject.toml so that we get the version number from versioneer for both the build artifacts and _version.py.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,8 +50,9 @@ sympy = ["sympy>=0.7.6.1", "scipy>=0.13.3", "jinja2>=2.10.1"]
 examples = ["jupyterlab", "matplotlib", "control>=0.9.3.post2,<=0.10.0"]
 all = ["pymoca[casadi,lxml,sympy,examples]"]
 
-[tool.setuptools.dynamic]
-version = { attr = "pymoca.__version__" }
+# TODO: Uncomment this when moving away from versioneer.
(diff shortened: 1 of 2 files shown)


kentrutan:
Next time I need to be more rigorous in testing build artifacts before releasing. The proper versions in the build artifact file names fooled me into thinking it worked. It was using versions from setuptools but giving the same format as versioneer if tagged and not dirty.

kentrutan:
I would like to push a 0.11.1 tag for release after merging this.

kentrutan:
I think adding a test of the generated versions for both development and deployment would be good since we use the dynamic versioning foot gun. I don't think I'll have time to work on it this week, so probably go ahead and release this to get it fixed. I think the added tests could go beyond just version checks, and the tests themselves could be a bit tricky to get right for both local and GitHub, so wouldn't want to rush through it. Also would probably only want to enable them for deployment. Worth looking around to see what others do for this.

jackvreeken (approved):
Yep, it's all a bit weird this dynamic versioning... Getting _something_ working is more important right now than the prettiest solution (if it even exists), so LGTM

PR #345, opened 2025-01-10
Fix build of maintenance/0.9.x

jackvreeken:
Unfortunately I still have a need for Pymoca 0.9.* , and need to move to Python 3.12+ because of platform constraints that are out of my hands.

Hopefully (and definitely once #307 lands) will be able to move RTC-Tools to a more recent Pymoca version.


jackvreeken:
Any general comments on this @kentrutan ?

kentrutan:
Sorry, don't have my laptop this weekend. Any general questions or can it wait till next week?

jackvreeken:
For sure can wait until next week, no rush :)

tools/create_conda_env.sh
```diff
@@ -12,7 +12,7 @@ conda update -q conda
 # Useful for debugging any issues with conda
 conda info -a
 conda config --add channels conda-forge
-conda create -n pymoca python=3.5 || echo "environment already created"
+conda create -n pymoca python=3.7 || echo "environment already created"
```
kentrutan: I get the error below with Python 3.7. Changing to 3.9 fixes it. Seems like it should be 3.9 anyway since that is the min version with this PR.

Feel free to ignore this since this script is not included in the distribution, but I figure I would bring it up since the commit had the 3.5 -> 3.7 change.

```
PackagesNotFoundError: The following packages are not available from current channels:

  - python=3.7*

Current channels:

  - https://conda.anaconda.org/conda-forge
  - defaults
```

tools/create_conda_env.sh
```diff
@@ -12,7 +12,7 @@ conda update -q conda
...
 conda install -n pymoca gcc jinja2 matplotlib numpy scipy sympy
```
kentrutan: After changing to Python 3.9, I get the error below. Removing gcc from the installed packages fixes it. Without gcc, the test suite passes and "python -m build" works fine, so I don't know what might have depended on it.

```
Channels:
 - conda-forge
 - defaults
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - gcc

```

pyproject.toml
```diff
@@ -2,6 +2,58 @@
...
+
+[project.urls]
+homepage = "https://github.com/pymoca/pymoca"
+
+[project.optional-dependencies]
+# Note that we need setuptools for the distutils.ccompiler dependency
+# in the CasADi backend.
+casadi = ["casadi>=3.4.0", "setuptools>=60.0.0"]
+lxml = ["lxml>=3.5.0", "scipy>=0.13.3"]
+sympy = ["sympy>=0.7.6.1", "scipy>=0.13.3", "jinja2>=2.10.1"]
```
kentrutan: Commit ae3e226fc72a8db2e7e4a847b2c5d5a89faf08b1 message:
> There is one test in the SymPy backend that fails with these changes, so
> we skip the test as we do not care about that backend in this
> maintenance branch.

Would pinning max sympy as in the earlier commit on this branch fa63d48 also work instead of skipping the test?

tools/create_conda_env.sh
```diff
@@ -12,7 +12,7 @@ conda update -q conda
...
 source activate pymoca
 # Note: casadi installed using pip since conda version currently
```
kentrutan, line 18: gcc is possibly needed for slycot which has to be built from source, but I only found that out after fixing the typo "eneduser" on line 24 😄. I'm now sorry for wasting time on it.

kentrutan (approved):
I checked out the .dev0 package on PyPI and it looks good as does  importing out of an editable install in a local repo. Just have a few minor comments and then should be good to go IMHO.

Probably should remove "Closes #338" in the commit message for 11ccc0f since the commit on master is actually what closed the issue.

Do you want to add antlr_build.py from 329bc181751747c5e78fb076e4550711ac71e99b and e5027d7cb6658492b045bd051f0414e025c84b54 since it was removed from setup.py? Also possibly move it into the antlr directory per e34e2ad1cbf710284bbfcd027bbf825f87392679 to be consistent with master? Also fix the distribution ala ccbe8f6da8f3462e15f8ab0fd334f00cdd6f931b if doing the others?

Also see the code comments. I don't think any of these are worth your time if you are in a hurry to get it done, so I'm going to go ahead and finish the review as approved.

tools/create_conda_env.sh
jackvreeken: Fixed!

tools/create_conda_env.sh
jackvreeken: Fixed

pyproject.toml
jackvreeken: I'll test it.

EDIT:
Yep, fixes it. I'll include in the list of commits I'm including changes of 😅

tools/create_conda_env.sh
jackvreeken: I'll just leave it at removing gcc. The conda script is not something I use personally (never have), and I'm not even sure it works at all anymore.

jackvreeken:
> Probably should remove "Closes https://github.com/pymoca/pymoca/issues/338" in the commit message for https://github.com/pymoca/pymoca/commit/11ccc0faf5d2d9ede6184d7a6f4e21f83e6cae57 since the commit on master is actually what closed the issue.

Done .

> antlr stuff

Also included in the `setup.py/pyproject.toml: Move mostly to pyproject.toml` commit now. Good catch!

jackvreeken:
Tagged an 0.9.2rc1 . I'll wait a week or two before tagging the final version. First I'll try using it by referring to it explicitly, tell others, and see if myself or they run any issues with it. I'll merge this one for now though. Those new potential issues can be their own PR.

Thanks for the review!

PR #352, opened 2026-03-02
sympy: Fix runtime integrated function shape

### LINKED ISSUE #351: CI Fails in backends.sympy.runtime with integrated function ValueError
CI pipeline has been failing since January 14 with the following exception generated inside `scipy.integrate`:

```
ValueError: Function to integrate must not return a tuple.
```

kentrutan:
Change the shape of the array that the function to integrate returns from (n,1) to a flat (n,) shape so that it is compatible with `scipy.integrate.ode`.

There was a change in SciPy 1.17.0 that caused all the sympy tests to fail. There were two changes to integration in that version, one or both of which likely caused the issue.

Closes #351

### DIFF
--- a/src/pymoca/backends/sympy/runtime.py
+++ b/src/pymoca/backends/sympy/runtime.py
@@ -112,7 +112,7 @@ class OdeModel:
         ss_subs.update(self.c0)
 
         # create f (dynamics) lambda function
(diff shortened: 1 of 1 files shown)


jackvreeken (approved):
Nice catch! And this seems to work all the way back to very old versions of SymPy as well.

PR #354, opened 2026-07-28
Add release notes and packaging testability

kentrutan:
For the upcoming release with a lot of new content, it would be good to add a summary to help understand the major user-facing improvements. At this stage of Pymoca's maturity, it would be also be good to keep maintenance lightweight and forgiving for maintainers. 

This adds a `CHANGELOG.md` file, but only containing the summary for each release and a link to GitHub if you want more detail. Upon push of a release tag, release notes are automatically assembled by the GitHub CI workflow `changelog` job with the summary for the release extracted from `CHANGELOG.md` followed by a categorized list of pull requests and a link to all the commits. The workflow will error out before any PyPI upload if the summary is not there for the given tag version.

Release tags ending in .dev* result in an upload to TestPyPI. Tags ending in rc* (non-dotted per PEP 440) are uploaded to PyPI, but are marked pre-release so they don't replace the current production release. These two options allow for testing of the packaging before official release. As before, bare version tags are uploaded to PyPI as a production release.

Along with the CI/CD updates above, we also upgrade security to using PyPI Trusted Publishing (OpenID Connect) with `pypa/gh-action-pypi-publish` instead of twine with credentials.

Finally, in the spirit of keeping maintenance lean, I noticed that the copyright end year was stale in the `LICENSE` file and changed it to just be the initial commit year, following the BSD license convention, to remove the need for future maintenance.

EDIT: Updated to PEP 440 compliant versioning per review comments.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,36 @@
+# Configuration for GitHub's automatically generated release notes.
+# https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes
+#
+# Merged PRs are sorted into the sections below by their labels. A PR lands in
+# the first category whose labels it matches, so order matters. The final
+# "Other Changes" catch-all ("*") ensures an unlabeled PR is still listed.
+changelog:
+  exclude:
+    labels:
+      - ignore
+    authors:
+      - dependabot
+  categories:
+    - title: Breaking Changes
+      labels:
+        - breaking
+    - title: Features
+      labels:
+        - enhancement
+        - feature
+    - title: Bug Fixes
+      labels:
+        - bug
+        - fix
+    - title: Performance
+      labels:
+        - performance
+    - title: Documentation
+      labels:
+        - documentation
+    - title: Dependencies
+      labels:
+        - dependencies
+    - title: Other Changes
+      labels:
+        - "*"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ on:
     # Every Wednesday at 6:06 UTC
     - cron: "6 6 * * 3"
 
+permissions:
+  contents: read
+
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
@@ -85,7 +88,7 @@ jobs:
           files: ./coverage.xml
           token: ${{ secrets.CODECOV_TOKEN }}
 
-  deploy:
+  build:
     needs: test
     runs-on: ubuntu-latest
     if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags') }}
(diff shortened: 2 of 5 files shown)


kentrutan:
**TL;DR:** This is a well-designed PR — the CHANGELOG-as-source-of-truth approach is clean: the "final edit" happens in a reviewable, versioned file *before* the tag, and the `changelog` job hard-fails the pipeline if the summary is missing. Nothing blocks merge; findings below are edge cases and setup items.

## What I verified works

- **The awk extraction is correct and robust.** I ran it against the actual `CHANGELOG.md` for tags `0.12.0`, `0.12.0.dev1`, and `0.12.0.rc2` — all correctly extract the `## 0.12.0` section; a tag with no section (`0.13.0`) correctly fails. The trailing-character guard properly prevents `0.12` from matching a `## 0.12.0` heading, and using `index()` (literal match) instead of a regex avoids metacharacter surprises in version strings. Passing the tag via `env: REF_NAME` rather than interpolating `${{ }}` into the script is the right injection-safe pattern.
- **Artifact plumbing is correct.** `dist` uploads/downloads to `dist/` (matching `gh-action-pypi-publish`'s default packages dir and `files: dist/*`), and `release-summary` with no `path` lands `release-summary.md` in the workspace root, matching `body_path`.
- **Gating is sound.** Both deploy jobs `need` `[build, changelog]`, `build` needs `test`, and skipped/failed needs skip the dependents — so no PyPI upload can happen on a tag with missing notes or failing tests, and non-tag events can never reach the deploy jobs even though their `if` only checks the tag name.
- **Trusted Publishing setup** (OIDC `id-token: write`, per-job environments, dropping the `PYPI_PASSWORD` twine flow) follows current PyPA best practice.

## Findings

**1. Tag-format trap: `0.12.0rc1` (PEP 440 canonical, no dot) slips past both `.rc` checks.** `contains(github.ref_name, '.rc')` and the `%%.rc*` stripping only handle the dotted form, so a canonical-form rc tag would route to the *production* `deploy` job with `prerelease: false`. The saving grace — confirmed by test — is that the changelog lookup for `## 0.12.0rc1` fails, so the pipeline stops before anything publishes. Still, the error message ("No release notes section for 0.12.0rc1") would be confusing. Suggest either also stripping/matching the undotted `rc`/`dev` forms, or simply documenting "always use dotted `.devN` / `.rcN` tag forms" in the `CHANGELOG.md` comment block, which already explains the tag conventions.

**2. rc releases may truncate the final release's auto-generated PR list.** `generate_release_notes` with no explicit `previous_tag_name` lets GitHub auto-pick the comparison baseline. After publishing `0.12.0.rc1` (a GitHub prerelease), the final `0.12.0` notes may be generated as "changes since rc1" — a near-empty PR list, with the real content stranded in the rc release. Worth a dry run to see what GitHub actually picks; if it misbehaves, the fixes are deleting rc releases before final tagging, or computing and passing the previous *stable* tag explicitly to the action.

**3. LICENSE year change: verify the 2014 claim.** I couldn't check from a shallow clone. The old text said `2016-2021`, which hints the project may date to 2016, though the pymola era could push it back to 2014. Worth a quick `git log --reverse --format='%ad' | head -1` in a full clone before merging, since an initial-commit-year convention is only self-maintaining if the year is right.

**4. Labels referenced in `release.yml` need to exist, and note where this PR itself will land.** The `ignore` exclude label and `breaking`, `feature`, `fix`, `performance` aren't GitHub defaults — create them or the categories stay empty. Also, this PR is labeled `ci-cd` + `documentation`; since `ci-cd` isn't in any category, it will file under **Documentation** (first match wins). If CI work should be separated in the notes, add a category keyed on `ci-cd` — otherwise it's fine as is.

**5. Pre-merge setup checklist (not code changes).** Before the first tag: create the `pypi` and `testpypi` GitHub environments, and register trusted publishers on both PyPI and TestPyPI (repo `pymoca/pymoca`, workflow `ci.yml`, matching environment names). Bonus: adding a required reviewer on the `pypi` environment would give a manual "approve before publish" pause — a nice complement to the pre-tag CHANGELOG edit, and it would also hold rc uploads.

**6. Nit: consider a top-level `permissions: contents: read`.** The deploy jobs correctly declare their own minimal permissions, but the other jobs run with the workflow default token scope. A one-line workflow-level restriction is cheap hardening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0126WPGnQw2XGd8u4mG4tpgA

---
_Generated by [Claude Code](https://claude.ai/code)_

kentrutan:
I plan to address items 1, 2, 3 and 6 from the Fable review above. The others are done or intentional. Will force-push when done.

kentrutan:
Items 1, 3, and 6 are fixed with the latest push. I verified that GitHub correctly creates the release notes with the full content with `gh api repos/pymoca/pymoca/releases/generate-notes -f tag_name=0.12.0`. 

I'll push dev and rc tags on the branch to verify everything is still working.

kentrutan:
Found an issue on the TestPyPI trial. Turns out PEP 440 convention is dotted dev and non-dotted rc. Talk about confusing. Pushed a fix to reject non-canonical to help our future selves when we undoubtedly forget 😆

kentrutan:
@jackvreeken would be great to have your 👍 or 👎 on the releasing/packaging changes.

